PR #5, opened 2017-07-05
adding retestAll option

owolabileg:
@awshi2 please review

### DIFF
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
@@ -35,12 +35,13 @@ public class RunMojo extends DiffMojo {
     protected boolean updateRunChecksums;
 
     /**
-     * Set this to "true" to run all tests, not just the affected ones. This is useful
+     * Set this option to "true" to run all tests, not just the affected ones. This option is useful
      * in cases where one is interested to measure the time to run all tests, while at the
-     * same time measuring the times for RTS and the number of tests that it selects.
-     * Note: Run with "-DstartsLogging=FINER" or "-DstartsLogging=FINEST" so that the
-     * "selected-tests" file, which contains the list of tests that would be run if this option
-     * is set to false, will be written to file.
+     * same time measuring the times for analyzing what tests to select and reporting the number of
+     * tests it would select.
+     * Note: Run with "-DstartsLogging=FINER" or "-DstartsLogging=FINEST" so that the "selected-tests"
+     * file, which contains the list of tests that would be run if this option is set to false, will
+     * be written to disk.
      */
     @Parameter(property = "retestAll", defaultValue = "false")
     protected boolean retestAll;
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
@@ -4,6 +4,7 @@
 
 package edu.illinois.starts.jdeps;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -33,6 +34,17 @@
     @Parameter(property = "updateRunChecksums", defaultValue = "true")
     protected boolean updateRunChecksums;
 
+    /**
+     * Set this to "true" to run all tests, not just the affected ones. This is useful
+     * in cases where one is interested to measure the time to run all tests, while at the
+     * same time measuring the times for RTS and the number of tests that it selects.
+     * Note: Run with "-DstartsLogging=FINER" or "-DstartsLogging=FINEST" so that the
```
august782: Isn't this note a little redundant, since these options would need to be set for even with retestAll is false in order to get the file "selected-tests"? Or is retestAll special in that regards?

owolabileg: I specified this here to preempt questions like, "how can I see what tests were selected, even though I run with retestAll?"

august782: I would reformat the text
"same time measuring the times for RTS and the number of tests that it select"
to something like 
"same time measuring the times for analyzing what tests to select and reporting the number of tests it would select"
because right now it sounds kind of like it measures only the time concerning running the selected tests.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -33,6 +34,17 @@
     @Parameter(property = "updateRunChecksums", defaultValue = "true")
     protected boolean updateRunChecksums;
 
+    /**
+     * Set this to "true" to run all tests, not just the affected ones. This is useful
```
alexgyori: naked this. set this -> set retestAll to be true

alexgyori: This is useful (naked this) -> This argument is useful... (otherwise the test may be useful... yes I'm doing a lot of writing these days:).

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -33,6 +34,17 @@
...
+     * in cases where one is interested to measure the time to run all tests, while at the
+     * same time measuring the times for RTS and the number of tests that it selects.
+     * Note: Run with "-DstartsLogging=FINER" or "-DstartsLogging=FINEST" so that the
+     * "selected-tests" file, which contains the list of tests that would be run if this option
```
alexgyori: such that the "selected-tests" file... will be written to the file?
The sentence is broken :).

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
owolabileg: Fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
owolabileg: Fixed.

owolabileg: Fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
owolabileg: Fixed.

PR #7, opened 2017-07-18
Fix cases with empty jdeps output

owolabileg:
When jdeps incrementally builds its cache, sometimes it finds `.jar` files with no `.class` files. This PR will allow jdeps write an empty `.graph` file in those cases.

### DIFF
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
@@ -79,7 +79,7 @@ public class RTSUtil {
         StringWriter output = new StringWriter();
         LOGGER.log(Level.FINE, "JDEPS ARGS:" + args);
         Main.run(args.toArray(new String[0]), new PrintWriter(output));
-        return getDepsFromJdepsOutput(output);
+        return output.getBuffer().length() != 0 ? getDepsFromJdepsOutput(output): new HashMap<String, Set<String>>();
     }
 
     public static Map<String, Set<String>> getDepsFromJdepsOutput(StringWriter jdepsOutput) {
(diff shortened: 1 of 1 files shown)


starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
```diff
@@ -79,7 +79,7 @@ public static void startProcess(ProcessBuilder pb) {
         StringWriter output = new StringWriter();
         LOGGER.log(Level.FINE, "JDEPS ARGS:" + args);
         Main.run(args.toArray(new String[0]), new PrintWriter(output));
-        return getDepsFromJdepsOutput(output);
+        return output.getBuffer().length() != 0 ? getDepsFromJdepsOutput(output) : new HashMap<String, Set<String>>();
```
august782: If there is no output, doesn't getDepsFromJdepsOutput just return the empty HashMap it creates in the beginning without adding anything into it? Is there some exception that gets thrown somewhere?

owolabileg: Yes, we get an `ArrayIndexOutOfBoundsException` on [Line 94](https://github.com/TestingResearchIllinois/starts/blob/21f5ebfe87e7d6918d8d4b3bbffff39fc368b3ac/starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java#L94) because when `toString()` is invoked on a `StringWriter` with `getBuffer().length == 0`, the result is a wierd, non-empty, blank string which still causes the body of the for loop to be entered.

august782: Put a comment about how the output can be empty due to JARs with no class files, and then merge.

august782: Also, don't forget to double-check the length of the line, for checkstyle.

owolabileg: Thanks for the reminder. I will add comment and merge.

owolabileg: BTW, Travis is activated now.

PR #8, opened 2017-08-30
Fix clz

owolabileg:
This PR fixes `-DdepFormat=ZLC` that was previously broken.

@august782 Please review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,4 @@
 language: java
 
-os:
-    - linux
-
-before_install:
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then jdk_switcher use oraclejdk8; fi
-
-env:
-    - CMD="mvn clean verify -B"
-
-script: eval $CMD
+script:
+    - mvn clean verify -B
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ before_install:
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then jdk_switcher use oraclejdk8; fi
 
 env:
-    - CMD="mvn verify -B"
+    - CMD="mvn clean verify -B"
 
 script: eval $CMD
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,12 @@
 language: java
 
+os:
+    - linux
+
 before_install:
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then jdk_switcher use oraclejdk8; fi
+
+env:
+    - CMD="mvn verify -B"
+
+script: eval $CMD
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: java
 
+before_install:
+    - jdk_switcher use oraclejdk8
+
 script:
     - mvn clean verify -B
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
@@ -16,16 +16,21 @@ import java.util.Set;
 import java.util.logging.Level;
 
 import edu.illinois.starts.util.Logger;
+import edu.illinois.starts.util.Pair;
 import org.ekstazi.check.AffectedChecker;
 
 /**
  * Utility methods for interacting with Ekstazi.
  */
 public class EkstaziHelper {
-    private static final Logger LOGGER = Logger.getGlobal();
+    public static final Logger LOGGER = Logger.getGlobal();
+    public static String notFirstRunMarker = "not-first-run.clz";
 
-    public static Set<String> getNonAffectedTests(String artifactsDir) {
+    public static Pair<Set<String>, Set<String>> getNonAffectedTests(String artifactsDir) {
         long start = System.currentTimeMillis();
+        if (isFirstRun(artifactsDir)) {
+            return null;
+        }
         ByteArrayOutputStream baosOut = new ByteArrayOutputStream();
         ByteArrayOutputStream baosErr = new ByteArrayOutputStream();
         PrintStream psOut = new PrintStream(baosOut);
@@ -42,7 +47,15 @@ public class EkstaziHelper {
         writeEkstaziDebugInfo(baosErr, artifactsDir);
         long end = System.currentTimeMillis();
         LOGGER.log(Level.FINEST, "[TIME]COMPUTING NON-AFFECTED: " + (end - start) + "ms");
-        return new HashSet<>(Arrays.asList(baosOut.toString().split("\n")));
+        Set<String> nonAffected = new HashSet<>(Arrays.asList(baosOut.toString().split("\n")));
+        //TODO: parse .clz files to find what changed in the same pass as finding nonaffected tests
+        Set<String> changed = new HashSet<>();
+        return new Pair<>(nonAffected, changed);
+    }
+
+    private static boolean isFirstRun(String artifactsDir) {
+        // If the notFirstRunMarker file does not exist, this is a first run
+        return !(new File(artifactsDir, notFirstRunMarker).exists());
     }
 
     public static Set<String> getNonAffectedTests(File basedir) {
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
@@ -73,7 +73,7 @@ public class EkstaziHelper {
      *
      * @param baosErr      Ekstazi Debug Output
      * @param artifactsDir Directory in which we store STARTS artifacts (i.e., ".starts")
-     * @return             The (possibly empty) list of changed classes
+     * @return             The (possibly empty) set of changed classes
      */
     private static Set<String> processEkstaziDebugInfo(ByteArrayOutputStream baosErr, String artifactsDir) {
         Set<String> changed = new HashSet<>();
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
@@ -25,6 +25,7 @@ import org.ekstazi.check.AffectedChecker;
 public class EkstaziHelper {
     public static final Logger LOGGER = Logger.getGlobal();
     public static String notFirstRunMarker = "not-first-run.clz";
+    public static String lineSeparator = System.getProperty("line.separator");
 
     public static Pair<Set<String>, Set<String>> getNonAffectedTests(String artifactsDir) {
         long start = System.currentTimeMillis();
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -1,4 +1,12 @@
...
+os:
+    - linux
+
 before_install:
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then jdk_switcher use oraclejdk8; fi
\ No newline at end of file
+    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then jdk_switcher use oraclejdk8; fi
+
+env:
+    - CMD="mvn clean verify -B"
```
august782: Why not just use "mvn verify -B" under script?

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
```diff
@@ -39,41 +45,48 @@
...
-        for (String line : Arrays.asList(baosErr.toString().split("\n"))) {
-            if (line.contains("::Diff::")) {
-                // This assumes that only classes in the application can
-                // change, and not classes in jars
-                String fqn = Writer.toFQN(line);
-                changed.add(fqn);
+        for (String line : Arrays.asList(baosErr.toString().split(lineSeparator))) {
+            String ekstaziDiffMarker = "::Diff:: ";
+            if (line.contains(ekstaziDiffMarker)) {
+                changed.add(line.split(ekstaziDiffMarker)[1]);
```
august782: Before changed was adding the toFQN of the line, while this new one takes the second half of ekstaziDiffMarker. Does this lead to a difference?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -99,8 +96,8 @@ protected void updateForNextRun(Set<String> nonAffected) throws MojoExecutionExc
         Set<String> affectedTests = new HashSet<>(allTests);
         affectedTests.removeAll(nonAffected);
         DirectedGraph<String> graph = null;
-        Map<String, Set<String>> testDeps = null;
-        Set<String> unreached = null;
+        Map<String, Set<String>> testDeps;
```
august782: Is it necessary to even define these variables here as they are not used in this scope afterwards?

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
```diff
@@ -39,41 +45,48 @@
...
+    public static Pair<Set<String>, Set<String>> getNonAffectedTests(File basedir) {
         long start = System.currentTimeMillis();
-        List<String> nonAffected = AffectedChecker.findNonAffectedClasses(basedir, getRootDirOption(basedir));
+        List<String> nonAffectedFiles = AffectedChecker.findNonAffectedClasses(basedir, getRootDirOption(basedir));
         long end = System.currentTimeMillis();
+        Set<String> nonAffectedTests = toFQN(new HashSet<>(nonAffectedFiles));
+        Set<String> changed = new HashSet<>();
         LOGGER.log(Level.FINEST, "[TIME]COMPUTING NON-AFFECTED(2): " + (end - start) + "ms");
-        return toFQN(new HashSet<>(nonAffected));
+        return new Pair<>(nonAffectedTests, changed);
```
august782: To clarify, changed will always be empty, right?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -69,25 +70,21 @@ public void execute() throws MojoExecutionException {
...
-            //TODO: parse .clz files to find what changed in the same pass as finding nonaffected tests
-            Set<String> changed = new HashSet<>();
-            //TODO: updateForNextRun, depending on updateChecksums
-            return new Pair<>(nonAffected, changed);
+            data = EkstaziHelper.getNonAffectedTests(getArtifactsDir());
         }
-        return null;
+        Set<String> changed = data == null ? new HashSet<String>() : data.getValue();
+        if (Logger.getGlobal().getLoggingLevel().intValue() <= Level.FINEST.intValue()) {
+            Writer.writeToFile(changed, "changed-classes", getArtifactsDir());
```
august782: In the case of CLZ, there would be empty changed, and so this just prints nothing, right?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -69,25 +70,21 @@ public void execute() throws MojoExecutionException {
...
-            Pair<Set<String>, Set<String>> data = zlcHelper.getChangedData(getArtifactsDir(), cleanBytes);
-            Set<String> changed = data == null ? new HashSet<String>() : data.getValue();
-            if (Logger.getGlobal().getLoggingLevel().intValue() <= Level.FINEST.intValue()) {
-                Writer.writeToFile(changed, "changed-classes", getArtifactsDir());
-            }
-            long end = System.currentTimeMillis();
-            Logger.getGlobal().log(Level.FINE, "[PROFILE] COMPUTING CHANGES: " + Writer.millsToSeconds(end - start));
-            return data;
+            data = zlcHelper.getChangedData(getArtifactsDir(), cleanBytes);
         } else if (depFormat == DependencyFormat.CLZ) {
```
august782: There are only two formats, ZLC and CLZ, right? If some other format makes it through (not sure if it's possible), the current logic just results in some default, empty returns, right? Should an exception be used instead, indicating some big break?

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
owolabileg: In this overloaded method, yes. Calling `AffectedChecker.findNonAffectedClasses()` does not return the changed classes.

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
owolabileg: Yes, we now store the `URL.toExternalForm()` to be consistent with what we store with `-DdepFormat=ZLC`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
owolabileg: Currently, if you run with a format that is not in `{CLZ,ZLC}`, e.g., `TTT`, you get an `IllegalArgumentException` exception: `Cannot convert 'TTT' to Enum: No enum constant edu.illinois.starts.enums.DependencyFormat.TTT`; that `depFormat` argument is backed by an `enum`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
owolabileg: There are two overloaded methods in `EkstaziHelper`; the one we are using does return a non-empty `changed` if run in the right logging level and the ekstazi config (`.ekstazirc`) is set correctly.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
owolabileg: I fixed that. Thanks!

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
```diff
@@ -39,41 +45,48 @@
...
+    }
+
+    private static boolean isFirstRun(String artifactsDir) {
+        // If the notFirstRunMarker file does not exist, this is a first run
+        return !(new File(artifactsDir, notFirstRunMarker).exists());
     }
 
-    private static void writeEkstaziDebugInfo(ByteArrayOutputStream baosErr, String artifactsDir) {
+    private static Set<String> writeEkstaziDebugInfo(ByteArrayOutputStream baosErr, String artifactsDir) {
```
august782: Name is a bit non-indicative of it returning changed classes. Maybe either refactor the name of the method or add a quick comment with the method as a reminder.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -172,6 +172,9 @@ public static String toFQNPlain(String line) {
     public static List<String> fqnsToExcludePath(Collection<String> fqns) {
         List<String> paths = new ArrayList<>();
         for (String fqn : fqns ) {
```
august782: Cut whitespace after fqns

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -69,25 +70,21 @@ public void execute() throws MojoExecutionException {
...
-            Set<String> nonAffected = EkstaziHelper.getNonAffectedTests(getArtifactsDir());
-            //TODO: parse .clz files to find what changed in the same pass as finding nonaffected tests
-            Set<String> changed = new HashSet<>();
-            //TODO: updateForNextRun, depending on updateChecksums
-            return new Pair<>(nonAffected, changed);
+            data = EkstaziHelper.getNonAffectedTests(getArtifactsDir());
```
august782: For CLZ, it is unnecessary to pass around the cleanBytes flag as an option?

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
owolabileg: changed name and added comment

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
owolabileg: Done. I also added `<module name="ParenPad"/>` to `checkstyle.xml` so we should catch this in the future.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
owolabileg: Unfortunately, yes. AFAIK, we have to control the `cleanBytes` for Ekstazi via the `hash.without.debuginfo = true` setting in the `.ekstazirc` file.

owolabileg:
@august782 Let me know if you have more comments on this PR, or whether I should merge now.

starts-core/src/main/java/edu/illinois/starts/helpers/EkstaziHelper.java
```diff
@@ -39,41 +45,56 @@
...
     }
 
-    private static void writeEkstaziDebugInfo(ByteArrayOutputStream baosErr, String artifactsDir) {
+    /**
+     * Process the Ekstazi debug output to get what classes Ekstazi thinks changed, and write those changed classes
+     * to file.
+     *
+     * @param baosErr      Ekstazi Debug Output
+     * @param artifactsDir Directory in which we store STARTS artifacts (i.e., ".starts")
+     * @return             The (possibly empty) list of changed classes
```
august782: It's actually a set

owolabileg: fixed

PR #9, opened 2017-08-31
Pull up yasgl dependency

DrRacket:
yasgl is used in both modules. Moving this dependency to root pom.
A side question is should we remove the jgrapht dependency since we are using yasgl now?


owolabileg:
Thanks! Please leave in jgrapht for now.

PR #13, opened 2017-09-19
adding functionality for retesting all when the classpath changes

milicahtanovic:
@owolabileg

### DIFF
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
@@ -6,11 +6,16 @@ package edu.illinois.starts.helpers;
 
 import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -67,6 +72,22 @@ public class Writer {
         String outFilename = artifactsDir + File.separator + "sf-classpath";
         try (BufferedWriter writer = getWriter(outFilename)) {
             writer.write(sfPathString + System.lineSeparator());
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+    }
+
+    public static void writeJarChecksums(String sfPathString, String artifactsDir) {
+        String outFilename = artifactsDir + File.separator + "jar-checksums";
+        try (BufferedWriter writer = getWriter(outFilename)) {
+            String[] jarsList = sfPathString.split(":");
+            for (int i = 0; i < jarsList.length; i++) {
+                if (jarsList[i].isEmpty()) {
+                    continue;
+                }
+                String mapping = getJarToChecksumMapping(jarsList[i]);
+                writer.write(mapping + "\n");
+            }
         } catch (IOException ioe) {
             ioe.printStackTrace();
         }
@@ -189,5 +210,28 @@ public class Writer {
 
     public static String millsToSeconds(long value) {
         return String.format("%.03f", (double) value / 1000.0);
+    }
+
+    public static String getJarToChecksumMapping(String jar) {
+        StringBuilder sb = new StringBuilder();
+        byte[] bytes = new byte[8192];
+        try {
+            MessageDigest md = MessageDigest.getInstance("MD5");
+            InputStream is = Files.newInputStream(Paths.get(jar));
+            DigestInputStream dis = new DigestInputStream(is, md);
+            while (dis.read(bytes) != -1) {
+                ;
+            }
+
+            byte[] mdbytes = md.digest();
+            for (int i = 0; i < mdbytes.length; i++) {
+                sb.append(Integer.toString((mdbytes[i] & 0xff) + 0x100 , 16).substring(1));
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        } catch (NoSuchAlgorithmException nsae) {
+            nsae.printStackTrace();
+        }
+        return jar + "," + sb.toString();
     }
 }
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
@@ -4,10 +4,20 @@
 
 package edu.illinois.starts.jdeps;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 
@@ -71,8 +81,11 @@ public class RunMojo extends DiffMojo {
     }
 
     protected void run(List<String> excludePaths) throws MojoExecutionException {
-        if (retestAll) {
+        String sfPathString = cleanClassPath(Writer.pathToString(getSureFireClassPath().getClassPath()));
+        if (retestAll || !compareClassPaths(sfPathString) || !checkJarChecksums(sfPathString)) {
             dynamicallyUpdateExcludes(new ArrayList<String>());
+            Writer.writeClassPath(sfPathString, artifactsDir);
+            Writer.writeJarChecksums(cleanClassPath(sfPathString), artifactsDir);
         } else {
             dynamicallyUpdateExcludes(excludePaths);
         }
@@ -100,5 +113,59 @@ public class RunMojo extends DiffMojo {
         Pair<Set<String>, Set<String>> data = computeChangeData();
         nonAffectedTests = data == null ? new HashSet<String>() : data.getKey();
         changedClasses  = data == null ? new HashSet<String>() : data.getValue();
+    }
+
+    private boolean compareClassPaths(String sfPathString) throws MojoExecutionException {
+        String oldSfPathFileName = getArtifactsDir() + "sf-classpath";
+        try {
+            String cleanOldSfPathFileName = cleanClassPath(Files.readAllLines(Paths.get(oldSfPathFileName)).get(0));
+            if (!sfPathString.equals(cleanOldSfPathFileName)) {
+                return false;
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+        return true;
+    }
+
+    private boolean checkJarChecksums(String cleanSfClassPath) throws MojoExecutionException {
+        String oldChecksumPathFileName = getArtifactsDir() + "jar-checksums";
+        Map<String, String> checksumMap = new HashMap<>();
+        try (BufferedReader fileReader = new BufferedReader(new FileReader(oldChecksumPathFileName))) {
+            String line;
+            while ((line = fileReader.readLine()) != null) {
+                String[] elems = line.split(",");
+                checksumMap.put(elems[0], elems[1]);
+            }
+            String[] jars = cleanSfClassPath.split(":");
+            for (int i = 0; i < jars.length; i++) {
+                String[] elems = Writer.getJarToChecksumMapping(jars[i]).split(",");
+                String oldCS = checksumMap.get(elems[0]);
+                if (!elems[1].equals(oldCS)) {
+                    return false;
+                }
+            }
+        } catch (IOException ioe) {
+            // log that the file doesn't exist, typically on the first run or after clean
+            ioe.printStackTrace();
+        }
+        return true;
+    }
+
+    private String cleanClassPath(String cp) {
+        String[] paths = cp.split(":");
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
+            if (paths[i].contains("/target/classes") || paths[i].contains("/target/test-classes")
+                || paths[i].contains("-SNAPSHOT.jar")) {
+                continue;
+            }
+            if (sb.length() == 0) {
+                sb.append(paths[i]);
+            } else {
+                sb.append(":" + paths[i]);
+            }
+        }
+        return sb.toString();
     }
 }


starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -72,6 +77,22 @@ public static void writeClassPath(String sfPathString, String artifactsDir) {
...
+    public static void writeJarChecksums(String sfPathString, String artifactsDir) {
+        String outFilename = artifactsDir + File.separator + "jar-checksums";
+        try (BufferedWriter writer = getWriter(outFilename)) {
+            String[] jarsList = sfPathString.split(":");
+            for (int i = 0; i < jarsList.length; i++) {
+                if (jarsList[i].isEmpty()) {
+                    continue;
+                }
+                String mapping = getJarToChecksumMapping(jarsList[i]);
+                writer.write(mapping + "\n");
```
august782: Use System.lineSeparator() instead of \n

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -190,4 +211,27 @@ public static void writeToLog(Set<String> set, String title, Logger logger) {
...
+
+    public static String getJarToChecksumMapping(String jar) {
+        StringBuilder sb = new StringBuilder();
+        byte[] bytes = new byte[8192];
+        try {
+            MessageDigest md = MessageDigest.getInstance("MD5");
+            InputStream is = Files.newInputStream(Paths.get(jar));
+            DigestInputStream dis = new DigestInputStream(is, md);
+            while (dis.read(bytes) != -1) {
+                ;
```
august782: What's with the empty loop?

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -190,4 +211,27 @@ public static void writeToLog(Set<String> set, String title, Logger logger) {
...
+            }
+
+            byte[] mdbytes = md.digest();
+            for (int i = 0; i < mdbytes.length; i++) {
+                sb.append(Integer.toString((mdbytes[i] & 0xff) + 0x100 , 16).substring(1));
```
august782: What is the goal of this math with hex values when making the String? Some comment could help.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -190,4 +211,27 @@ public static void writeToLog(Set<String> set, String title, Logger logger) {
...
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        } catch (NoSuchAlgorithmException nsae) {
+            nsae.printStackTrace();
+        }
+        return jar + "," + sb.toString();
```
august782: If you're already using a StringBuilder, just use the StringBuilder to concatenate the Strings, instead of using the + (which will make a new StringBuilder anyway).

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
+            if (paths[i].contains("/target/classes") || paths[i].contains("/target/test-classes")
+                || paths[i].contains("-SNAPSHOT.jar")) {
+                continue;
+            }
+            if (sb.length() == 0) {
+                sb.append(paths[i]);
+            } else {
+                sb.append(":" + paths[i]);
```
august782: If you're already using a StringBuilder, it's best to just use the StringBuilder methods to concatenate Strings instead of using +.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            // log that the file doesn't exist, typically on the first run or after clean
+            ioe.printStackTrace();
+        }
+        return true;
+    }
+
+    private String cleanClassPath(String cp) {
+        String[] paths = cp.split(":");
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
```
august782: Is it ever possible for the paths to be in a different order? Is it necessary to sort first as to make the String always deterministic?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
     }
+
+    private boolean compareClassPaths(String sfPathString) throws MojoExecutionException {
+        String oldSfPathFileName = getArtifactsDir() + "sf-classpath";
+        try {
+            String cleanOldSfPathFileName = cleanClassPath(Files.readAllLines(Paths.get(oldSfPathFileName)).get(0));
+            if (!sfPathString.equals(cleanOldSfPathFileName)) {
+                return false;
+            }
+        } catch (IOException ioe) {
```
august782: So upon IOException the return is true, meaning the classpaths would be considered the same. Is this what should happen?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
         nonAffectedTests = data == null ? new HashSet<String>() : data.getKey();
         changedClasses  = data == null ? new HashSet<String>() : data.getValue();
     }
+
+    private boolean compareClassPaths(String sfPathString) throws MojoExecutionException {
```
august782: Maybe change the name to instead of compareClassPaths to be something like checkIfSameClassPath, just to get the feel that the method returns a boolean.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            if (!sfPathString.equals(cleanOldSfPathFileName)) {
+                return false;
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+        return true;
+    }
+
+    private boolean checkJarChecksums(String cleanSfClassPath) throws MojoExecutionException {
```
august782: Similar comment about name as for compareClassPaths.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            }
+            String[] jars = cleanSfClassPath.split(":");
+            for (int i = 0; i < jars.length; i++) {
+                String[] elems = Writer.getJarToChecksumMapping(jars[i]).split(",");
+                String oldCS = checksumMap.get(elems[0]);
+                if (!elems[1].equals(oldCS)) {
+                    return false;
+                }
+            }
+        } catch (IOException ioe) {
```
august782: Similar comment concerning exception and return as in compareClassPaths.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -72,6 +77,22 @@ public static void writeClassPath(String sfPathString, String artifactsDir) {
         }
     }
 
+    public static void writeJarChecksums(String sfPathString, String artifactsDir) {
+        String outFilename = artifactsDir + File.separator + "jar-checksums";
```
alexgyori: It's safer to not concatenate paths this way (e.g., artifactsDir may or may not end with a File.separator already). There is a method Paths.get(...).

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Ok, fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: It's not empty technically, it is filling the 'bytes' array.
I used 'while (dis.read(bytes) != -1);' but checkstyle complained, it requires {}.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Yes, I added the description now. It is computing the checksum of the jars.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -72,6 +77,22 @@ public static void writeClassPath(String sfPathString, String artifactsDir) {
...
+        try (BufferedWriter writer = getWriter(outFilename)) {
+            String[] jarsList = sfPathString.split(":");
```
alexgyori: In unix systems the path separator is : but in windows it is ; (semicolon). To be portable use File.pathSeparator which gives you the right thing. As a rule of thumb, whenever you find yourself using some string literal ask if that's portable. If you set it yourself, it's fine; if it comes from the platform, check.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -72,6 +77,22 @@ public static void writeClassPath(String sfPathString, String artifactsDir) {
...
+            for (int i = 0; i < jarsList.length; i++) {
```
alexgyori: You could use an enhanced for loop here to be more java-ic (is that the equivalent of pythonic).

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
alexgyori: it is filling and then overwritting right? If the loop executes more than once you only get the last 8192 bytes and loose the prior ones, which may or may not be what you want.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+    private boolean checkJarChecksums(String cleanSfClassPath) throws MojoExecutionException {
+        String oldChecksumPathFileName = getArtifactsDir() + "jar-checksums";
+        Map<String, String> checksumMap = new HashMap<>();
+        try (BufferedReader fileReader = new BufferedReader(new FileReader(oldChecksumPathFileName))) {
+            String line;
+            while ((line = fileReader.readLine()) != null) {
+                String[] elems = line.split(",");
+                checksumMap.put(elems[0], elems[1]);
+            }
+            String[] jars = cleanSfClassPath.split(":");
```
alexgyori: use pathSeparator if needed...

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Ok, fixed it.
Yes, it could be empty and it would just write an empty line. Now I added a condition to skip if it could not generate the checksum.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            }
+        } catch (IOException ioe) {
+            // log that the file doesn't exist, typically on the first run or after clean
+            ioe.printStackTrace();
+        }
+        return true;
+    }
+
+    private String cleanClassPath(String cp) {
+        String[] paths = cp.split(":");
```
alexgyori: same as above about the pathSeparator

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
alexgyori: enhanced for loop to be more java-ic.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
+            if (paths[i].contains("/target/classes") || paths[i].contains("/target/test-classes")
```
alexgyori: this is definitely not portable; in windows you'd have \\.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,58 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+                || paths[i].contains("-SNAPSHOT.jar")) {
```
alexgyori: endsWith instead of contains maybe?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: I agree, done.

alexgyori: To explain why these comments keep coming up, read this: http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html
It's a performance hit if you concatenate a lot of stuff... probably you don't but it's good to write good code.

alexgyori: Do you want to use pathSeparator here to be portable? i.e., it's : on unix and ; on windows?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Good point. I will do set intersection instead?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Yes.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Changed it.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: I changed it to return true in the try block and false otherwise.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Updated that too.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Ok, used the same fix.

milicahtanovic: Another thing with these exceptions is that they will happen on the first run always. We may want to silence them in that case?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
owolabileg: If you are already putting them in sets just check for equality with `equals()`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
owolabileg: check for file existence before the try:

```java
if (!new File(oldChecksumPathFileName).exists()) {
     return false;
}
```

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: I understand. We are going to fix it and the whole algorithm in fact.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: We are using Paths.get(...) later in the code for this exact string.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Ok, done.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: True, but it will eventually be broken down to this so it's not that important.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Ok, changed it.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Ok, done.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Added the condition.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Yes, I will consider that.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
alexgyori: ok... but is there a reason to not use it here?

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
alexgyori: Not done here.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -72,6 +77,25 @@ public static void writeClassPath(String sfPathString, String artifactsDir) {
...
+                if (jarsList[i].isEmpty()) {
+                    continue;
+                }
+                String mapping = getJarToChecksumMapping(jarsList[i]);
+                if (mapping.split(",").length < 2) {
+                    continue;
+                }
+                writer.write(mapping + System.lineSeparator());
```
alexgyori: have two calls to write rather than concatenate (which creates a new StringBuilder...)

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
         nonAffectedTests = data == null ? new HashSet<String>() : data.getKey();
         changedClasses  = data == null ? new HashSet<String>() : data.getValue();
     }
+
+    private boolean checkIfSameClassPath(String sfPathString) throws MojoExecutionException {
+        String oldSfPathFileName = getArtifactsDir() + "sf-classpath";
```
alexgyori: similarly here you are assuming that the path returned by the getArtifactsDir ends with / or \. If it doesn't this path is incorrect. Use Paths.get(...) maybe.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            ioe.printStackTrace();
+        }
+        return noException;
+    }
+
+    private String cleanClassPath(String cp) {
+        String[] paths = cp.split(File.pathSeparator);
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
+            if (paths[i].contains("/target/classes") || paths[i].contains("/target/test-classes")
```
alexgyori: These paths don't quite work on Windows.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Ok, didn't see that. Changed it now.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: I just added File.separator.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Ok, I used the File.separator-s here too.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: That would require to overload the method to which this string is passed and we don't want to do that.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Done.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: It is fixed now.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: This is changed to use commons-codec to get hex.

owolabileg:
@august782 @denizarsan @alexgyori I'd like to merge this ASAP. Please take one final look.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
august782: I think what Alex is saying is that you would never be sure that getArtifactsDir() actually returns a String that ends with File.separator; later somebody could change it, or someone reading this code would never be sure that it ends with File.separator (and would need to jump 2 or 3 files just to double-check, which I did...). Just to be safe, it would be good to explicitly use something like Paths.get(...). As of now, it does indeed work though, since getArtifactsDir() does end with File.separator.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -117,12 +117,14 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
 
     private boolean checkIfSameClassPath(String sfPathString) throws MojoExecutionException {
         String oldSfPathFileName = getArtifactsDir() + "sf-classpath";
-        Set<String> sfClassPathSet = new HashSet<>(Arrays.asList(sfPathString.split(":")));
+        Set<String> sfClassPathSet = new HashSet<>(Arrays.asList(sfPathString.split(File.pathSeparator)));
+        if (!new File(oldSfPathFileName).exists()) {
```
august782: Maybe move this conditional check one line earlier, before setting sfClassPathSet, as if it does not exist, making that Set is not necessary.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -117,12 +117,14 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            return false;
+        }
         try {
             String cleanOldSfPathFileName = cleanClassPath(Files.readAllLines(Paths.get(oldSfPathFileName)).get(0));
             Set<String> oldSfClassPathSet = new HashSet<>(Arrays.asList(cleanOldSfPathFileName.split(":")));
```
august782: Same comment as before about : versus File.pathSeparator.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -135,13 +137,16 @@ private boolean checkIfSameJarChecksums(String cleanSfClassPath) throws MojoExec
         String oldChecksumPathFileName = getArtifactsDir() + "jar-checksums";
         Map<String, String> checksumMap = new HashMap<>();
         boolean noException = true;
+        if (!new File(oldChecksumPathFileName).exists()) {
```
august782: Same comment as before, moving conditional closer to where it should be checked to exit earlier.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Done.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Already fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Fixed in the newer version.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Moved.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
         nonAffectedTests = data == null ? new HashSet<String>() : data.getKey();
         changedClasses  = data == null ? new HashSet<String>() : data.getValue();
     }
+
+    private boolean checkIfSameClassPath(String sfPathString) throws MojoExecutionException {
+        String oldSfPathFileName = Paths.get(getArtifactsDir() + File.separator + "sf-classpath");
```
august782: If you are using Paths.get(...), you then don't need to append File.separator, you can just use Paths.get(getArtifactsDir(), "sf-classpath"), with the parts as arguments.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+        if (!new File(oldSfPathFileName).exists()) {
+            return false;
+        }
+        try {
+            String cleanOldSfPathFileName = cleanClassPath(Files.readAllLines(Paths.get(oldSfPathFileName)).get(0));
+            Set<String> sfClassPathSet = new HashSet<>(Arrays.asList(sfPathString.split(File.pathSeparator)));
+            Set<String> oldSfClassPathSet = new HashSet<>(Arrays.asList(cleanOldSfPathFileName.split(File.pathSeparator)));
+            if (oldSfClassPathSet.equals(oldSfClassPathSet)) {
```
august782: This is checking if oldSfClassPathSet is equal to itself. I believe you need it to be some other variable?

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -72,6 +78,26 @@ public static void writeClassPath(String sfPathString, String artifactsDir) {
...
+    public static void writeJarChecksums(String sfPathString, String artifactsDir) {
+        String outFilename = artifactsDir + File.separator + "jar-checksums";
+        try (BufferedWriter writer = getWriter(outFilename)) {
+            String[] jarsList = sfPathString.split(File.pathSeparator);
+            for (int i = 0; i < jarsList.length; i++) {
+                if (jarsList[i].isEmpty()) {
+                    continue;
+                }
+                String mapping = getJarToChecksumMapping(jarsList[i]);
+                if (mapping.split(",").length < 2) {
```
denizarsan: Where does "2" come from?

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -190,4 +216,35 @@ public static void writeToLog(Set<String> set, String title, Logger logger) {
...
+    /**
+     * Compute the checksum for the given map and return the jar
+     * and the checksum as a string.
+     *
+     * @param jar         The jar whose checksum we need to compute.
+     */
+    public static String getJarToChecksumMapping(String jar) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(jar);
+        sb.append(",");
```
denizarsan: You can chain these two appends for readability.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
         nonAffectedTests = data == null ? new HashSet<String>() : data.getKey();
         changedClasses  = data == null ? new HashSet<String>() : data.getValue();
     }
+
+    private boolean checkIfSameClassPath(String sfPathString) throws MojoExecutionException {
```
denizarsan: What do you think about naming this "isSameClassPath" instead?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            if (oldSfClassPathSet.equals(oldSfClassPathSet)) {
+                return true;
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+        return false;
+    }
+
+    private boolean checkIfSameJarChecksums(String cleanSfClassPath) throws MojoExecutionException {
```
denizarsan: Same as before: "hasSameJarChecksum"

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+                }
+            }
+        } catch (IOException ioe) {
+            noException = false;
+            ioe.printStackTrace();
+        }
+        return noException;
+    }
+
+    private String cleanClassPath(String cp) {
```
denizarsan: Same: "getCleanClassPath"

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+        String[] paths = cp.split(File.pathSeparator);
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
+            if (paths[i].contains(File.separator + "target" +  File.separator + "classes")
+                || paths[i].contains(File.separator + "target" + File.separator + "test-classes")
+                || paths[i].contains("-SNAPSHOT.jar")) {
```
denizarsan: contains -> endsWith

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+                continue;
+            }
+            if (sb.length() == 0) {
+                sb.append(paths[i]);
+            } else {
+                sb.append(File.pathSeparator);
+                sb.append(paths[i]);
```
denizarsan: You can chain these two appends for readability.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+            ioe.printStackTrace();
+        }
+        return noException;
+    }
+
+    private String cleanClassPath(String cp) {
+        String[] paths = cp.split(File.pathSeparator);
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < paths.length; i++) {
+            if (paths[i].contains(File.separator + "target" +  File.separator + "classes")
```
august782: Small optimization, it may be good to create the one String that represents File.separator + "target" + ... outside the loop, so there is no String concatenation on every iteration of the loop.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
august782: I think it would still be good to use Paths.get(...), for same reasons as in other places with this pattern.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
august782: By chaining, do you mean something like sb.append(jar + ",")? I would actually not recommend that, since underneath that would create a new StringBuilder just to concatenate the two Strings, and since we already have a StringBuilder going it is a bit unoptimal.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
august782: Same comment about comment about chaining from before.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -101,4 +114,69 @@ protected void setChangedAndNonaffected() throws MojoExecutionException {
...
+                return true;
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+        return false;
+    }
+
+    private boolean checkIfSameJarChecksums(String cleanSfClassPath) throws MojoExecutionException {
+        String oldChecksumPathFileName = getArtifactsDir() + File.separator + "jar-checksums";
```
august782: Same comment concerning using Paths.get(...).

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
denizarsan: No. I mean sb.append(jar).append(“,”).

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
denizarsan: Same reply: I mean `sb.append(foo).append(bar)`. Sorry if it wasn’t clear.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Yeah, I wasn't gonna include this in the commit. I am not sure I want to use Path.get(...). I just need a string here. I will remove Path.get(...).

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Sounds good.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
august782: You will actually need to put .toString() at the end to make it work for the String. I would still recommend it since Paths.get(...) has the logic for determining whether or not to use the File.separator when connecting the parts, leading to a cleaner String.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Oh I fix that on the other branch. Good catch.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: I taught that the problem was the path separator? 
What is the actual suggestion?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Ok.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
august782: The suggestion is to do "String oldChecksumPathFileName = Paths.get(getArtifactsDir(), "jar-checksums").toString();"
The reason is because it is unclear (without spending some time looking) whether or not getArtifactsDir() has the File.separator at the end or not. If it does, it leads to a kind of ugly String with the double /. While this might not break functionality, it is usually better to use some utility methods that provide the benefit of the checking.

milicahtanovic: Fixed.

alexgyori: what happens if you have:

"/home/milicah/" + File.separator + "something"?
You will have two /, i.e., //, in the middle of the string. Which may or may not break things depending on what you do. The first string may or may not have / at the end, you either put an assertion that ensures the first string doesn't end with /, or use Paths.get everywhere which has the logic to handle all these edge cases which we come up with because we actually ran into them.

You can see how `Paths.get` is implemented  by digging into: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/nio/file/Paths.java#84

Alex

milicahtanovic: I actually had that at some point. It didn't brake anything, but yeah it could. 
I changed this to use Paths.get(...).

milicahtanovic:
@alexgyori @august782 @denizarsan @owolabileg All the comments have been addressed, I think this branch is ready to merge.

denizarsan:
@milicahtanovic You have some conflicts. Please resolve them before merging.

august782:
I don't think all comments have been addressed yet. Please at least make sure you have responded to each comment and answer any questions or acknowledge the comment.

milicahtanovic:
Oh, I didn't see them. For some reason they didn't show up on the discussion board. I will go over this, sorry.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: This is how I save them when I write to the file -> <jar>,<checksum>.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Yeah, I considered that but I don't think it makes much difference.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
denizarsan: Can you add a comment explaining this? It will make things easier for others or for you in the future.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Done.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Done.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Ok, done.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
milicahtanovic: Done.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Yes, but as I said in this case it requires changing a method that is widely used within the whole project. This string gets passed to Paths.get(...) anyways.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
milicahtanovic: Added.

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
august782: I don't quite understand what widely used method is being changed here. Is it not enough to just change artifactDir + File.separator + "jar-checksum" to Paths.get(artifactDir, "jar-checksum").toString()?

milicahtanovic: ok.

august782:
Please confirm on some small examples that the functionality is still working, that our suggestions did not break anything; try modifying the classpath and JARs on the classpath to test. After that, I believe it should be ready to merge.

milicahtanovic:
I tested with changing the dependency versions and it behaves as expected.

PR #15, opened 2017-09-29
optionally distinguish newly-added classes from impacted ones

owolabileg:


starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -190,4 +190,13 @@ public static void writeToLog(Set<String> set, String title, Logger logger) {
...
         return String.format("%.03f", (double) value / 1000.0);
     }
+
+    /**
+     * Convert the result of URL.toExternalForm() on classes in the program to a fully-qualified name.
+     * @param url External form of the URL to convert
+     * @return A fully-qualified name of the URL
+     */
+    public static String urlToFQN(String url) {
+        return url.split("classes" + File.separator)[1].replace(".class", "").replace(File.separator, ".");
```
august782: There's the small thing where it's possible that "classes/" shows up somewhere else later in the path, but seems extremely unlikely...

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -180,4 +180,27 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
...
         return new HashSet<>(Arrays.asList(tests.split(",")));
     }
+
+    public static Set<String> getExistingClasses(String artifactsDir) {
+        Set<String> existingClasses = new HashSet<>();
+        long start = System.currentTimeMillis();
+        File zlc = new File(artifactsDir, zlcFile);
+        if (!zlc.exists()) {
+            LOGGER.log(Level.FINEST, "@NoExistingZLCFile. First Run?");
+            return null;
```
august782: Is it necessary to return null? Maybe just return the empty set, and then there is no need for a null check on use.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -180,4 +180,27 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
...
+            for (String line : zlcLines) {
+                if (line.startsWith("file")) {
+                    existingClasses.add(Writer.urlToFQN(line.split(" ")[0]));
+                }
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+        long end = System.currentTimeMillis();
+        LOGGER.log(Level.FINEST, "[TIME]COMPUTING NEWLY-ADDED CLASSES: " + (end - start) + "ms");
```
august782: Change the logging message to indicate that it is computing existing classes.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
```diff
@@ -76,14 +92,16 @@ private void updateForNextRun(List<String> allClasses) throws MojoExecutionExcep
         ZLCHelper zlcHelper = new ZLCHelper();
         zlcHelper.updateZLCFile(result.getTestDeps(), loader, getArtifactsDir(), new HashSet<String>());
         long end = System.currentTimeMillis();
+        if (logger.getLoggingLevel().intValue() == Level.FINER.intValue()) {
+            Writer.writeClassPath(sfPathString, artifactsDir);
```
august782: Before, it was using artifactsDir, a passed in argument into the save() method, but it was originally the result of getArtifactsDir() in this context. Maybe it should be the same here?

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
owolabileg: Added that assumption

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
owolabileg: returning empty set now

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
owolabileg: done

starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
owolabileg: done

starts-core/src/main/java/edu/illinois/starts/helpers/Writer.java
```diff
@@ -190,4 +190,14 @@ public static void writeToLog(Set<String> set, String title, Logger logger) {
...
         return String.format("%.03f", (double) value / 1000.0);
     }
+
+    /**
+     * Convert the result of URL.toExternalForm() on classes in the program to a fully-qualified name.
+     * @param url External form of the URL to convert
+     * @return A fully-qualified name of the URL
+     */
+    public static String urlToFQN(String url) {
+        //ASSUMPTION: "classes/" rarely occurs in the rest of the path
```
august782: //ASSUMPTION: -> // ASSUMPTION:

owolabileg: Done

PR #18, opened 2017-11-05
write changed-classes to file at level FINER

yujinz:


owolabileg:
I believe that what you need for your purpose is to write `nonAffected` to file, not `changed`. Please do that in the appropriate part of the code.

PR #19, opened 2017-11-06
 write nonAffected to disk optionally

yujinz:


starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -82,6 +82,10 @@ public void execute() throws MojoExecutionException {
         if (Logger.getGlobal().getLoggingLevel().intValue() <= Level.FINEST.intValue()) {
             Writer.writeToFile(changed, "changed-classes", getArtifactsDir());
         }
+        if (Logger.getGlobal().getLoggingLevel().intValue() == Level.FINER.intValue()) {
+            Set<String> nonAffected = data == null ? new HashSet<String>() : data.getKey();
+            Writer.writeToFile(nonAffected, "nonAffected-classes", getArtifactsDir());
+        }
```
owolabileg: Please extract this into a method that gets called at the end of `execute()`. The reasons: (1) the current location unnecessarily adds to the time for `computeChangeData()` (2) at the end of `execute()`, `nonAffected` is already resolved so there will be no reason to check again the nullness of `data`.

owolabileg:
Also, you'll need to remove your commit b1ff5e3, which was rejected in #18

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
yujinz: Will `execute()` of `DiffMojo` be called when we run `starts:starts`? I wrote the function there and tested it, but I think the function is not called.

testingsavvy: There should be a way to not write this file during `computeChangedData()`.

owolabileg: There is a way to not write this file during `computeChanged()`. Hint: what `execute` methods are invoked on `starts:starts`?

yujinz: It's `RunMojo`'s `execute()`. Line 59 of `RunMojo` already writes the file at FINEST, so I'm just going to submit a PR to change it to FINER.

yujinz:
I squashed the commits. Is it the correct way to remove the commit?

owolabileg:
Thanks a lot; this looks good! One last change before we merge: can you please add an option such that `nonAffected` is written to disk only when the option is `true` or `loggingLevel == FINEST`? The reason is that we already write a couple of files at `FINEST`, and I don't want users of your feature to be slowed down by the writing of those other files.

owolabileg:
Yes, squashing is fine.

yujinz:
Just making sure we are not writing when `loggingLevel == FINER` anymore?

owolabileg:
Yes, we want to write the file when `loggingLevel == FINEST || optionThatYouWillAdd`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -45,6 +45,15 @@
     @Parameter(property = "retestAll", defaultValue = "false")
     protected boolean retestAll;
 
+    /**
+     * Set this to "true" to save nonAffectedTests as a file on disk. This improve the time of
```
owolabileg: `save nonAffectedTests as a file on disk` -> `save nonAffectedTests to a file on disk`
`This improve the time of` -> `This improves the time for`
`for the second time` -> `twice`
`does the same thing` -> what same thing?

owolabileg:
@yujinz Once you address the minor fixes to comment, this should be ready to merge.

owolabileg:
@yujinz, should this be merged? I noticed the code here is also part of #35

yujinz:
Yes, please. I think my all solutions are going to utilize this parameter.

PR #20, opened 2017-11-07
Two-invocations solution of offline mode

yujinz:
This is an initial solution of the offline mode by letting the user run `mvn starts:starts -D updateRunChecksums=false` and then `mvn starts:update`.  It will be replaced by a parallel solution in the end, but I appreciate any comments on the current solution to help me keep on the right track.

`starts:update` is a goal I implemented for updating the dependency graph. It gets the nonAffected data by either running `computeChangedData()` again or load its result from disk if the `mvn starts:starts -D updateRunChecksums=false` has written it to disk.

Thank you!

### DIFF
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
@@ -45,6 +45,15 @@ public class RunMojo extends DiffMojo {
     @Parameter(property = "retestAll", defaultValue = "false")
     protected boolean retestAll;
 
+    /**
+     * Set this to "true" to save nonAffectedTests as a file on disk. This improve the time of
+     * updating test dependencies in offline mode by not running computeChangeData() for the second
+     * time.
+     * Note: Run with "-DstartsLogging=FINEST" does the same thing.
+     */
+    @Parameter(property = "writeNonAffected", defaultValue = "false")
+    protected boolean writeNonAffected;
+
     protected Set<String> nonAffectedTests;
     protected Set<String> changedClasses;
     private Logger logger;
@@ -56,7 +65,7 @@ public class RunMojo extends DiffMojo {
         setChangedAndNonaffected();
         List<String> excludePaths = Writer.fqnsToExcludePath(nonAffectedTests);
         setIncludesExcludes();
-        if (logger.getLoggingLevel().intValue() <= Level.FINER.intValue()) {
+        if (writeNonAffected || logger.getLoggingLevel().intValue() <= Level.FINEST.intValue()) {
             Writer.writeToFile(nonAffectedTests, "non-affected-tests", getArtifactsDir());
         }
         run(excludePaths);
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
@@ -56,7 +56,7 @@ public class RunMojo extends DiffMojo {
         setChangedAndNonaffected();
         List<String> excludePaths = Writer.fqnsToExcludePath(nonAffectedTests);
         setIncludesExcludes();
-        if (logger.getLoggingLevel().intValue() <= Level.FINEST.intValue()) {
+        if (logger.getLoggingLevel().intValue() <= Level.FINER.intValue()) {
             Writer.writeToFile(nonAffectedTests, "non-affected-tests", getArtifactsDir());
         }
         run(excludePaths);
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2015 - Present. The STARTS Team. All Rights Reserved.
+ */
+
+package edu.illinois.starts.jdeps;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
+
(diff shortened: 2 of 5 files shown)


starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+/**
+ * Update class and test dependencies.
+ */
+@Mojo(name = "update", requiresDirectInvocation = true, requiresDependencyResolution = ResolutionScope.TEST)
+public class UpdateMojo extends DiffMojo {
+    public void execute() throws MojoExecutionException {
+        long start = System.currentTimeMillis();
+        Set<String> nonAffected = null;
```
owolabileg: Do you really need to set `nonAffected` to null?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
```diff
@@ -0,0 +1,65 @@
...
+        String inFilename = getArtifactsDir() + File.separator + "non-affected-tests";
+        File inFile = new File(inFilename);
```
owolabileg: Please use better variable names than `inFile` and `inFilename`; we want variable names that say what the files are about.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
```diff
@@ -0,0 +1,65 @@
...
+        if (inFile.isFile()) {
```
owolabileg: you should check if `isFile() && exists()`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
```diff
@@ -0,0 +1,65 @@
...
+                    + Writer.millsToSeconds(end - start));
+        } else {
+            Pair<Set<String>, Set<String>> data = computeChangeData();
+            nonAffected = data == null ? new HashSet<String>() : data.getKey();
+            long end = System.currentTimeMillis();
+            Logger.getGlobal().log(Level.FINE, "[PROFILE] computeChangeData(): "
+                            + Writer.millsToSeconds(end - start));
+        }
+        //updateDiffChecksums is always true for this mojo
+        updateForNextRun(nonAffected);
```
owolabileg: I like what you have done here to always call `updateForNextRun`. However, for conformity to other Mojos, can you please add and `update*Checksums` option which is `true` by default?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+@Mojo(name = "update", requiresDirectInvocation = true, requiresDependencyResolution = ResolutionScope.TEST)
+public class UpdateMojo extends DiffMojo {
+    public void execute() throws MojoExecutionException {
+        long start = System.currentTimeMillis();
+        Set<String> nonAffected = null;
+        String inFilename = getArtifactsDir() + File.separator + "non-affected-tests";
+        File inFile = new File(inFilename);
+        if (inFile.isFile()) {
+            nonAffected = readFromFile(inFile, getArtifactsDir());
```
owolabileg: Replace method `readFromFile` with a one line call to `Files.readAllLines()`

starts-plugin/src/main/resources/META-INF/maven/lifecycle.xml
```diff
@@ -21,4 +21,19 @@
             </phase>
         </phases>
     </lifecycle>
+    <lifecycle>
+        <id>starts-graph</id>
+        <phases>
+            <phase>
+                <id>initialize</id>
```
owolabileg: what's the reason to put this in initialize phase?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -56,7 +56,7 @@ public void execute() throws MojoExecutionException {
         setChangedAndNonaffected();
         List<String> excludePaths = Writer.fqnsToExcludePath(nonAffectedTests);
         setIncludesExcludes();
-        if (logger.getLoggingLevel().intValue() <= Level.FINEST.intValue()) {
+        if (logger.getLoggingLevel().intValue() <= Level.FINER.intValue()) {
```
owolabileg: note the requested change in your other PR (#19)

starts-plugin/src/main/java/edu/illinois/starts/jdeps/UpdateMojo.java
yujinz: https://docs.oracle.com/javase/7/docs/api/java/io/File.html#isFile() says isFile() returns
"true if and only if the file denoted by this abstract pathname exists and is a normal file". Do we still want to check if `exists()`?

owolabileg: You are right. No need for `exists()`

starts-plugin/src/main/resources/META-INF/maven/lifecycle.xml
yujinz: As all phases before the one that contains the goal will be executed when we invoke the goal, putting the goal at the earliest phase will reduce the overhead. The `update` goal doesn't need to execute any maven phases since they haven been executed by the starts:starts command.
`initialize` is the second phase of maven, the first one is `validate` but I am not sure if I can put the goal there.

owolabileg:
@yujinz is this PR subsumed by #35?

yujinz:
Yes.

PR #26, opened 2017-11-14
fixes IT on Windows

jayfurrie:
Passes ``mvn install`` and ``mvn integration-test`` on both Windows 10 and macOS
left log messages in to help others see into what's happening - I can remove them if that's preferred.
Does anyone have a linux machine they can easily run this on?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,20 +1,6 @@
 version: '{build}'
 os: Windows Server 2012
 install:
-  - ps: |
-      Add-Type -AssemblyName System.IO.Compression.FileSystem
-      if (!(Test-Path -Path "C:\maven" )) {
-        (new-object System.Net.WebClient).DownloadFile(
-          'http://www.us.apache.org/dist/maven/maven-3/3.2.5/binaries/apache-maven-3.2.5-bin.zip',
-          'C:\maven-bin.zip'
-        )
-        [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\maven-bin.zip", "C:\maven")
-      }
-      
-  - cmd: SET M2_HOME=C:\maven\apache-maven-3.2.5
-  - cmd: SET PATH=%M2_HOME%\bin;%JAVA_HOME%\bin;%PATH%
-  - cmd: SET MAVEN_OPTS=-XX:MaxPermSize=2g -Xmx4g
-  - cmd: SET JAVA_OPTS=-XX:MaxPermSize=2g -Xmx4g
   - cmd: mvn --version
   - cmd: java -version
 build_script:


jayfurrie:
removed comments, reverted abstractmojointerceptor methods to be generic rather than hardcode "JavaAgent.class"

still passes ``mvn install`` and ``mvn integration-test`` on macOS and Windows.
Installing linux on VM to ensure stability

jayfurrie:
works on linux

jayfurrie:
windows CI: https://ci.appveyor.com/project/jayfurrie/starts-j4780
linux/macOS CI: https://travis-ci.org/jayfurrie/starts

.travis.yml
```diff
@@ -1,7 +1,22 @@
 language: java
 
-before_install:
-    - jdk_switcher use oraclejdk8
+#default depth is 50, unnecessary
```
august782: Let's not mess with the default depth, in case there are situations where multiple commits come in at once and when a build is triggered it might not then be able to check out the commit it needs if the depth is set too low.

.travis.yml
```diff
@@ -1,7 +1,22 @@
...
+    - os: osx
+      osx_image: xcode8
 
 script:
     - mvn clean verify -B
+
+cache:
+    directories:
+    - $HOME/.m2
+    - $HOME/.binaries
```
august782: What are .binaries? Do we generate these?

appveyor.yml
```diff
@@ -0,0 +1,26 @@
...
+        [System.IO.Compression.ZipFile]::ExtractToDirectory("C:\maven-bin.zip", "C:\maven")
+      }
+      
+  - cmd: SET M2_HOME=C:\maven\apache-maven-3.2.5
+  - cmd: SET PATH=%M2_HOME%\bin;%JAVA_HOME%\bin;%PATH%
+  - cmd: SET MAVEN_OPTS=-XX:MaxPermSize=2g -Xmx4g
+  - cmd: SET JAVA_OPTS=-XX:MaxPermSize=2g -Xmx4g
+  - cmd: mvn --version
+  - cmd: java -version
+build_script:
```
august782: I don't really know appveyor, but I assume these correspond to the install and script phases of Travis? Should they match the commands used in Travis?

triggerbuild
```diff
@@ -0,0 +1 @@
+modifying this to trigger build.
```
denizarsan: Please remove this file before merging

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
```diff
@@ -10,19 +10,21 @@
 /**
  * This class is duplicated from Ekstazi, with minor changes.
  */
+
 public final class AgentLoader {
     private static final String TOOLS_JAR_NAME = "tools.jar";
     private static final String CLASSES_JAR_NAME = "classes.jar";
     private static final String AGENT_INIT = AgentLoader.class.getName() + " Initialized";
 
+
```
denizarsan: extra space

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
```diff
@@ -10,19 +10,21 @@
 /**
  * This class is duplicated from Ekstazi, with minor changes.
  */
+
```
denizarsan: extra space

appveyor.yml
```diff
@@ -0,0 +1,26 @@
+version: '{build}'
+os: Windows Server 2012
+install:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
+      if (!(Test-Path -Path "C:\maven" )) {
```
august782: Just as a curiosity, does this mean Maven is not installed by default for Appveyor, so it needs to be donwloaded?

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
```diff
@@ -10,19 +10,21 @@
...
     public static boolean loadDynamicAgent() {
         try {
             if (System.getProperty(AGENT_INIT) != null) {
                 return true;
             }
             System.setProperty(AGENT_INIT, "");
-
-            URL agentJarURL = AbstractMojoInterceptor.extractJarURL(JavaAgent.class);
+            URL agentJarURL = AbstractMojoInterceptor.class.getResource("JavaAgent.class");
```
august782: Is this extra step of calling getResouce() first key to making things work on Windows?

.travis.yml
denizarsan: I agree that having a depth of 1 might introduce such problems but isn't 50 a bit too much though?

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
jayfurrie: yes
previously, we'd call AbstractMojoInterceptor's extractJarURL, with a passed in JavaAgent.class.
Then, we use JavaAgent.class to find the URL of JavaAgent.class - according to oracle docs, this is a bit redundant. 
The change I made is to use the AbstractMojoInterceptor.class's location to find JavaAgent.class's URL. Since the two are both in the same Jar, there is no additional stuff required - JavaAgent.class is directly available at the same file level as AbstractMojoInterceptor.class, no need to drill deeper into the jar with slashes and File.SeparatorChars.
The same is true on MacOS. This avoids using any AbstractMojoInterceptor class specific methods to find the URL of JavaAgent, since we can do so from a Class method instead (which AbstractMojoInterceptor can use since it's a Class)

.travis.yml
denizarsan: I know this was reverted to the default value, but I still want to hear what you think @august782

.travis.yml
```diff
@@ -1,7 +1,17 @@
...
+  
+matrix:
+  include:
+    - os: linux
+      jdk: oraclejdk8
+    - os: osx
+      osx_image: xcode8
 
 script:
     - mvn clean verify -B
```
august782: Just noticing that this command uses clean. Since the install phase that runs before should already build everything, we don't need to delete all the built binaries just to recompile them in the script phase.

appveyor.yml
```diff
@@ -0,0 +1,26 @@
...
+      }
+      
+  - cmd: SET M2_HOME=C:\maven\apache-maven-3.2.5
+  - cmd: SET PATH=%M2_HOME%\bin;%JAVA_HOME%\bin;%PATH%
+  - cmd: SET MAVEN_OPTS=-XX:MaxPermSize=2g -Xmx4g
+  - cmd: SET JAVA_OPTS=-XX:MaxPermSize=2g -Xmx4g
+  - cmd: mvn --version
+  - cmd: java -version
+build_script:
+  - mvn clean compile -B -DskipTest
```
august782: To better match what Travis is doing, we should use its default, which is "mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V". Also, let's remove the clean from the test_script phase as well (after making the change in the .travis.yml).

denizarsan:
You should squash the commits during merge as the commits contained in the PR might make the history confusing. There are a lot of commits with message "Update .travis.yml".

appveyor.yml
jayfurrie: it appears from blog posts back in 2015/2016, maven may not have been available.

It has since been added to the build system image. They use maven 3.3.9

https://www.appveyor.com/docs/build-environment/#java

.travis.yml
august782: I think it's fine to leave it as the default value as I don't believe checking out with depth 50 hurts performance significantly (and it's still much better than checking out the entire history).

appveyor.yml
august782: Then do we need this part here to download Maven, if it is provided by default?

.travis.yml
jayfurrie: done

appveyor.yml
jayfurrie: I think the problem is that we don't know for sure where maven is installed

If you scroll thru the link above, most of the other programs provided by the build environment have paths telling you where that particular program is available.

But no such thing exists for maven

And to ensure we know where maven is installed, we check to see if it is where we suspect it is (following the structure of the other apps provided by the build server).

And this is particularly important for when we want to cache, we need to know the path. Once we do it for the first time, since it's part of the cache, !(Test-Path -Path "C:\maven" )) should always return $TRUE so we'll skip the downloads

appveyor.yml
jayfurrie: i've changed these to match travis (after removing the clean from travis)

appveyor.yml
jayfurrie: done

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
august782: To make things a bit less confusing, use different variable names for the URL obtained from calling getResource() and the one from calling extractJarURL().

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
jayfurrie: fixed

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
jayfurrie: fixed

appveyor.yml
august782: From scrolling through that link, there are a lot of software that it says is installed but does not provide a link to where it exists on the VM. I assume you already tried not downloading anything and not setting anything extra to the path, and just invoking mvn crashes because mvn is not on the path?

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
```diff
@@ -34,12 +34,10 @@ public static boolean loadAgent(URL aju) throws Exception {
         if (toolsJarFile == null) {
             return false;
         }
-
```
august782: Why are these lines being deleted? As a rule of thumb, it's best to make only necessary changes for a PR. Refactoring changes, such as adding empty lines or deleting empty lines for readability, can be done in a separate PR just for that.

jayfurrie: deleted these thinking they were the ones I was asked to delete - whoops

august782: There are now extra spaces in the beginning on those lines; those should be removed.

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
```diff
@@ -21,9 +21,9 @@ public static boolean loadDynamicAgent() {
                 return true;
             }
             System.setProperty(AGENT_INIT, "");
-
-            URL agentJarURL = AbstractMojoInterceptor.extractJarURL(JavaAgent.class);
-            return loadAgent(agentJarURL);
+            URL agentJarURL = AbstractMojoInterceptor.class.getResource("JavaAgent.class");
+            agentJarURLConnection = AbstractMojoInterceptor.extractJarURL(agentJarURL);
```
august782: Is this agentJarURLConnection declared somewhere? Actually, if it isn't declared, should this not compile?

appveyor.yml
jayfurrie: checking this now

starts-core/src/main/java/edu/illinois/starts/maven/AgentLoader.java
jayfurrie: not sure how checkstyle passed before, but after adding these back in, checkstyle is erroring out

august782: There are extra spaces at the beginning of the lines added back in, so Checkstyle is picking up on those. We can see that from the diff view given on GitHub. Also, I would highly recommend building locally after your changes to make sure things pass before pushing, so such Checkstyle problems can be handled locally first.

PR #27, opened 2017-11-14
check null obect per EvoSuite test

peifengjing:
This corner case is found by EvoSuite tests. If the input moreEdges is a null object (which is not common but possible), the method addAll will fail.

There is another failure which may be considered as a fake failure when using DirectedGraph.fromGiraphString(). If the input file is a null object, it cannot be read or written.

### DIFF
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
@@ -46,7 +46,12 @@ public class Cache {
         // Some projects depend directly on jars in the standard library, so
         // we want to check there as well
         jarsInCache.addAll(checkMissingJarsInJDKCache(missing));
-        moreEdges.addAll(loadCachedEdges(jarsInCache));
+        if (moreEdges != null) {
+            moreEdges.addAll(loadCachedEdges(jarsInCache));
+        }
+        else {
+            moreEdges = new ArrayList<String>(loadCachedEdges(jarsInCache));
+        }
     }
 
     private HashSet<String> getJarsMissingFromCache(Set<String> jarsInCache) {
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
@@ -48,9 +48,8 @@ public class Cache {
         jarsInCache.addAll(checkMissingJarsInJDKCache(missing));
         if (moreEdges != null) {
             moreEdges.addAll(loadCachedEdges(jarsInCache));
-        }
-        else {
-            moreEdges = new ArrayList<String>(loadCachedEdges(jarsInCache));
+        } else {
+            moreEdges = new ArrayList<>(loadCachedEdges(jarsInCache));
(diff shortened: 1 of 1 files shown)


august782:
From my checking of the code, it seems the only use of the relevant method, loadM2EdgesFromCache(), is in edu.illinois.starts.jdeps.BaseMojo, in the method prepareForNextRun(). In this method, right before the call to loadM2EdgesFromCache(), the moreEdges passed in is instantiated as a new ArrayList. It does not seem possible for moreEdges to be null.

Can you share the test generated by EvoSuite which reveals this failure? Does it directly invoke loadM2EdgesFromCache() using a null object?

owolabileg:
@peifengjing, any updates on this?

peifengjing:
Sorry that I did not know why the notification went to my junk email box. You are right that the EvoSuite does not consider where the method is used in the program but only did generate unit test for the method. Here is the unit test:
`  @Test(timeout = 4000)
  public void test3()  throws Throwable  {
      MockFile mockFile0 = new MockFile("UZKym'/Fy[Y<Z", "UZKym'/Fy[Y<Z");
      EvoSuiteFile evoSuiteFile0 = new EvoSuiteFile("/Users/Peifeng/Desktop/Study/Topics-in-Software-Engineering/GitLab/evosuite/case_study/starts/starts-core/UZKym'/Fy[Y<Z/UZKym'/Fy[Y<Z/.graph");
      FileSystemHandling.createFolder(evoSuiteFile0);
      Cache cache0 = new Cache(mockFile0, "UZKym'/Fy[Y<Z");
      LinkedList<String> linkedList0 = null;
      cache0.loadM2EdgesFromCache(linkedList0, ".jar");
      assertEquals(0, linkedList0.size());
  }`

august782:
Does EvoSuite only generate unit tests for public methods? Can/should we restrict the visibility of the method to indicate that it is not usable by just anyone? Is there some annotation we should introduce to show that it should not be expecting null? Otherwise, it does not make sense to have to put a null-check for all arguments for every single method we have in the code.

peifengjing:
I can understand that it is meaningless to add null-check for every method and it is very annoying, and it is a very tricky corner case found by EvoSuite.

However, the problem comes from the design pattern. The method "loadM2EdgesFromCache()" is only utilized in class [BaseMojo line 227](https://github.com/TestingResearchIllinois/starts/blob/5967c25702b83f7620d956a8d5d6ac7051f54c83/starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java#L227), where the initialization of the input ArrayList is right before the method (line 223) and it is always empty when passing to the method. I am not sure why it is designed in this way, and it seems to makes the code complicated with defects (like this corner case for the method).

I am fine to close this PR without any change. Please let me know how do you think.

owolabileg:
@peifengjing I don't think a nullness check is necessary here as `moreEdges` is always initialized before passing it to `loadM2EdgesFromCache`.  You are welcome to make a different PR to make `loadM2EdgesFromCache` return a `List<String>` that is used to initialize `moreEdges`. Thanks! :+1:

PR #28, opened 2017-11-15
Add option for incremental dependency finding

hksskh:
Add option for incremental dependency finding in STARTS:
1) in prepareForNextRun function in BaseMojo, pass changedClasses and newClasses to Loadables so that jdeps only find dependencies for changed and new classes instead of all classes in localpath
2) only find transitiveClosures for affected tests instead of all tests
3) update ZLC file with dependencies for changed classes, and add dependencies for new classes, instead of creating new ZLC file each time
You can use 'mvn starts:starts -DincrementalUpdate=true' to enable incremental dependency finding. 'computeUnreached' is also made an option in DiffMojo.

### DIFF
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
@@ -12,17 +12,17 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 
+import edu.illinois.starts.constants.StartsConstants;
 import edu.illinois.starts.util.Logger;
 
 /** Utility methods for dealing with cached files. */
 
-public class Cache {
+public class Cache implements StartsConstants {
     private static final Logger LOGGER = Logger.getGlobal();
-    private static Set<String> newJarClasses;
+    private static final String GRAPH_EXTENSION = ".graph";
 
     File jdepsCache;
     String m2Repo;
@@ -32,7 +32,6 @@ public class Cache {
     public Cache(File jdepsCache, String m2Repo) {
         this.jdepsCache = jdepsCache;
         this.m2Repo = m2Repo;
-        newJarClasses = new HashSet<>();// only place to clear new jar classes. carefully call this
     }
 
     public void loadM2EdgesFromCache(List<String> moreEdges, String pathString) {
@@ -52,10 +51,6 @@ public class Cache {
         moreEdges.addAll(loadCachedEdges(jarsInCache));
     }
 
-    public static Set<String> getNewJarClasses() {
-        return newJarClasses;
-    }
-
     private HashSet<String> getJarsMissingFromCache(Set<String> jarsInCache) {
         HashSet<String> missing = new HashSet<>();
         if (!cpJars.equals(jarsInCache)) {
@@ -71,7 +66,7 @@ public class Cache {
         for (String jar : missing) {
             File missingFile = new File(jar);
             String fileName = missingFile.getName();
-            File jdkJarGraphFile = new File(jdepsCache, fileName.replace(".jar", ".graph"));
+            File jdkJarGraphFile = new File(jdepsCache, fileName.replace(JAR_EXTENSION, GRAPH_EXTENSION));
             if (jdkJarGraphFile.exists()) {
                 found.add(jdkJarGraphFile.getName());
             } else {
@@ -82,10 +77,7 @@ public class Cache {
         for (String jar : notFound) {
             //1. parse with jdeps and store in the cache
             List<String> args = new ArrayList<>(Arrays.asList("-v", jar));
-            Map<String, Set<String>> jdepsOutput = RTSUtil.runJdeps(args);
-            // add new jar classes
-            newJarClasses.addAll(jdepsOutput.keySet());
-            Writer.writeDepsToFile(jdepsOutput, createCacheFile(jar).getAbsolutePath());
+            Writer.writeDepsToFile(RTSUtil.runJdeps(args), createCacheFile(jar).getAbsolutePath());
             newlyCreated.add(jar);
         }
         //2. add newly-created graphs to list of jars that were previously found in cache
@@ -126,7 +118,7 @@ public class Cache {
     }
 
     private File createCacheFile(String jar) {
-        String cachePath = jar.replace(m2Repo + File.separator, "").replace(".jar", ".graph");
+        String cachePath = jar.replace(m2Repo + File.separator, EMPTY).replace(JAR_EXTENSION, GRAPH_EXTENSION);
         return new File(jdepsCache, cachePath);
     }
 
@@ -137,7 +129,7 @@ public class Cache {
         Set<String> jars = new HashSet<>();
         String[] splitCP = sfPathString.split(File.pathSeparator);
         for (int i = 0; i < splitCP.length; i++) {
-            if (splitCP[i].endsWith(".jar")) {
+            if (splitCP[i].endsWith(JAR_EXTENSION)) {
                 jars.add(splitCP[i]);
             }
         }
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 
@@ -21,6 +22,7 @@ import edu.illinois.starts.util.Logger;
 
 public class Cache {
     private static final Logger LOGGER = Logger.getGlobal();
+    private static Set<String> newJarClasses;
 
     File jdepsCache;
     String m2Repo;
@@ -30,6 +32,7 @@ public class Cache {
     public Cache(File jdepsCache, String m2Repo) {
         this.jdepsCache = jdepsCache;
         this.m2Repo = m2Repo;
+        newJarClasses = new HashSet<>();// only place to clear new jar classes. carefully call this
     }
 
     public void loadM2EdgesFromCache(List<String> moreEdges, String pathString) {
@@ -47,6 +50,10 @@ public class Cache {
         // we want to check there as well
         jarsInCache.addAll(checkMissingJarsInJDKCache(missing));
         moreEdges.addAll(loadCachedEdges(jarsInCache));
+    }
+
+    public static Set<String> getNewJarClasses() {
+        return newJarClasses;
     }
 
     private HashSet<String> getJarsMissingFromCache(Set<String> jarsInCache) {
@@ -75,7 +82,10 @@ public class Cache {
         for (String jar : notFound) {
             //1. parse with jdeps and store in the cache
             List<String> args = new ArrayList<>(Arrays.asList("-v", jar));
-            Writer.writeDepsToFile(RTSUtil.runJdeps(args), createCacheFile(jar).getAbsolutePath());
+            Map<String, Set<String>> jdepsOutput = RTSUtil.runJdeps(args);
+            // add new jar classes
+            newJarClasses.addAll(jdepsOutput.keySet());
+            Writer.writeDepsToFile(jdepsOutput, createCacheFile(jar).getAbsolutePath());
             newlyCreated.add(jar);
         }
         //2. add newly-created graphs to list of jars that were previously found in cache
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
@@ -30,20 +30,23 @@ public class Loadables {
 
     Map<String, Set<String>> deps;
     List<String> extraEdges;
-    private List<String> classesToAnalyze;
+    private List<String> changedClasses;
+    private List<String> newClasses;
+    private List<String> affectedTests;
+    private List<String> allTests;
     private File cache;
     private String sfPathString;
     private DirectedGraph<String> graph;
     private Map<String, Set<String>> transitiveClosure;
     private Set<String> unreached;
-    private Set<String> affectedTests;
     private boolean filterLib;
     private Classpath surefireClasspath;
     private String artifactsDir;
 
-    public Loadables(List<String> classesToAnalyze, String artifactsDir, String sfPathString,
+    public Loadables(List<String> changedClasses, List<String> newClasses, String artifactsDir, String sfPathString,
                      boolean filterLib, File cache) {
-        this.classesToAnalyze = classesToAnalyze;
+        this.changedClasses = changedClasses;
+        this.newClasses = newClasses;
         this.artifactsDir = artifactsDir;
         this.sfPathString = sfPathString;
         this.filterLib = filterLib;
@@ -77,20 +80,28 @@ public class Loadables {
     }
 
     public Loadables create(List<String> moreEdges, Classpath sfClassPath,
-                            boolean computeUnreached) {
+                            boolean computeUnreached, boolean incrementalUpdate) {
         setSurefireClasspath(sfClassPath);
         LOGGER.log(Level.FINEST, "More: " + moreEdges.size());
         extraEdges = moreEdges;
         long startTime = System.currentTimeMillis();
-        deps = getDepMap(sfPathString, classesToAnalyze);
+        deps = getDepMap(sfPathString, changedClasses, newClasses, incrementalUpdate);
         long jdepsTime = System.currentTimeMillis();
         graph = makeGraph(deps, extraEdges);
         long graphBuildingTime = System.currentTimeMillis();
-        transitiveClosure = getTransitiveClosurePerClass(graph, new ArrayList<>(affectedTests));
+        if (incrementalUpdate) {
+            transitiveClosure = getTransitiveClosurePerClass(graph, affectedTests);
+        } else {
+            transitiveClosure = getTransitiveClosurePerClass(graph, allTests);
+        }
         long transitiveClosureTime = System.currentTimeMillis();
         if (computeUnreached) {
-            //Todo: findUnreached may be affected by incrementally finding test dependencies
-            unreached = findUnreached(deps, transitiveClosure);
+            //TODO: findUnreached may be affected by incrementally finding test dependencies
+            Map<String, Set<String>> allTransitiveClosure = transitiveClosure;
+            if (incrementalUpdate) {
+                allTransitiveClosure = getTransitiveClosurePerClass(graph, allTests);
+            }
+            unreached = findUnreached(deps, allTransitiveClosure);
             LOGGER.log(Level.INFO, "UNREACHED(count): " + unreached.size());
         }
         long findUnreachedTime = System.currentTimeMillis();
@@ -159,9 +170,10 @@ public class Loadables {
         return builder;
     }
 
-    public Map<String, Set<String>> getDepMap(String pathToUse, List<String> classes)
-            throws IllegalArgumentException {
-        if (classes.isEmpty()) {
+    public Map<String, Set<String>> getDepMap(String pathToUse, List<String> changedClasses, List<String> newClasses,
+                                              boolean incrementalUpdate) throws IllegalArgumentException {
+        if ((incrementalUpdate && changedClasses.isEmpty() && newClasses.isEmpty())
+                || (!incrementalUpdate && allTests.isEmpty())) {
             //There are no test classes, no need to waste time with jdeps
             return null;
         }
@@ -184,17 +196,23 @@ public class Loadables {
             jdepsClassPath = Writer.pathToString(localPaths);
         }
         args.addAll(Arrays.asList("-cp", jdepsClassPath));
-        args.addAll(classes); // only let jdeps analyze dependencies for changed/new classes
-        //args.addAll(localPaths);
+        if (incrementalUpdate) {
+            args.addAll(changedClasses);
+            args.addAll(newClasses);
+        } else {
+            args.addAll(localPaths);
+        }
         LOGGER.log(Level.FINEST, "JDEPS CMD: " + args);
         Map<String, Set<String>> depMap = RTSUtil.runJdeps(args);
-        // load jdeps dependencies from last revision
-        RTSUtil.addDepsFromJdepsFile(depMap, new HashSet<>(classes),
-                artifactsDir + File.separator + "jdeps-out");
-        //if (LOGGER.getLoggingLevel().intValue() == Level.FINEST.intValue()) {
-        // store jdeps dependencies in this revision for future use
-        Writer.writeDepsToFile(depMap, artifactsDir + File.separator + "jdeps-out");
-        //}
+        if (incrementalUpdate) {
+            // load jdeps dependencies from last revision
+            RTSUtil.addDepsFromJdepsFile(depMap, new HashSet<>(changedClasses),
+                    new HashSet<>(newClasses), artifactsDir + File.separator + "jdeps-out");
+        }
+        if (LOGGER.getLoggingLevel().intValue() == Level.FINEST.intValue() || incrementalUpdate) {
+            // store jdeps dependencies in this revision for future use
+            Writer.writeDepsToFile(depMap, artifactsDir + File.separator + "jdeps-out");
+        }
         return depMap;
     }
 
@@ -227,7 +245,11 @@ public class Loadables {
         this.surefireClasspath = surefireClasspath;
     }
 
-    public void setAffectedTests(Set<String> affectedTests) {
+    public void setAffectedTests(List<String> affectedTests) {
         this.affectedTests = affectedTests;
+    }
+
+    public void setAllTests(List<String> allTests) {
+        this.allTests = allTests;
     }
 }
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
@@ -15,6 +15,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 
+import edu.illinois.starts.constants.StartsConstants;
 import edu.illinois.starts.util.ChecksumUtil;
 import edu.illinois.starts.util.Logger;
 import edu.illinois.yasgl.DirectedGraph;
@@ -25,7 +26,7 @@ import org.ekstazi.util.Types;
 /**
  * Utility methods for loading several things from disk.
  */
-public class Loadables {
+public class Loadables implements StartsConstants {
     private static final Logger LOGGER = Logger.getGlobal();
 
     Map<String, Set<String>> deps;
@@ -71,7 +72,7 @@ public class Loadables {
         List<String> localPaths = new ArrayList<>();
         if (surefireClasspath != null) {
             for (String path : surefireClasspath.getClassPath()) {
-                if (!path.endsWith(".jar") && new File(path).exists()) {
+                if (!path.endsWith(JAR_EXTENSION) && new File(path).exists()) {
                     localPaths.add(path);
                 }
             }
@@ -218,7 +219,7 @@ public class Loadables {
 
     private void addEdgesToGraphBuilder(DirectedGraphBuilder<String> builder, List<String> edges) {
         for (String edge : edges) {
-            String[] parts = edge.split(" ");
+            String[] parts = edge.split(WHITE_SPACE);
             if (parts.length != 2) {
                 LOGGER.log(Level.SEVERE, "@@BrokenEdge: " + edge);
                 continue;
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
@@ -36,6 +36,7 @@ public class Loadables {
     private DirectedGraph<String> graph;
     private Map<String, Set<String>> transitiveClosure;
     private Set<String> unreached;
+    private Set<String> affectedTests;
     private boolean filterLib;
     private Classpath surefireClasspath;
     private String artifactsDir;
@@ -85,9 +86,10 @@ public class Loadables {
         long jdepsTime = System.currentTimeMillis();
         graph = makeGraph(deps, extraEdges);
         long graphBuildingTime = System.currentTimeMillis();
-        transitiveClosure = getTransitiveClosurePerClass(graph, classesToAnalyze);
+        transitiveClosure = getTransitiveClosurePerClass(graph, new ArrayList<>(affectedTests));
         long transitiveClosureTime = System.currentTimeMillis();
         if (computeUnreached) {
+            //Todo: findUnreached may be affected by incrementally finding test dependencies
             unreached = findUnreached(deps, transitiveClosure);
             LOGGER.log(Level.INFO, "UNREACHED(count): " + unreached.size());
         }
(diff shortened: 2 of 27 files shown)


owolabileg:
@august782, @denizarsan, @milicahtanovic please help review this. It's a big PR, but @hksskh has agreed to make smaller, easier to review PRs in the future.

hksskh:
Nov. 20. 
Add two commits:
merge from STARTS master to integrate retestAll when jar changes
Add incrementalUpdate option to use in ImpactedMojo, and set up a standalone zlc file for ImpactedMojo

owolabileg:
@hksskh It's best if you don't add any more commits to this PR. Notice that no one else got the chance to review this yet, likely because it involves a lot of code changes at once. Thanks

starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
```diff
@@ -30,6 +32,7 @@
     public Cache(File jdepsCache, String m2Repo) {
         this.jdepsCache = jdepsCache;
         this.m2Repo = m2Repo;
+        newJarClasses = new HashSet<>();// only place to clear new jar classes. carefully call this
```
august782: It is kind of unclear that one needs to make a new Cache in order to clear the contents of a static variable. It would be better to make it a bit more transparent, perhaps with an explicit method.

On that note, it is also unclear why newJarClasses has to be static to begin with, especially if its creation is tied to creating a new Cache to start with. Can you provide some intuition for this?

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
```diff
@@ -76,19 +80,28 @@ public Loadables(List<String> classesToAnalyze, String artifactsDir, String sfPa
...
         extraEdges = moreEdges;
         long startTime = System.currentTimeMillis();
-        deps = getDepMap(sfPathString, classesToAnalyze);
+        deps = getDepMap(sfPathString, changedClasses, newClasses, incrementalUpdate);
         long jdepsTime = System.currentTimeMillis();
         graph = makeGraph(deps, extraEdges);
         long graphBuildingTime = System.currentTimeMillis();
-        transitiveClosure = getTransitiveClosurePerClass(graph, classesToAnalyze);
+        if (incrementalUpdate) {
+            transitiveClosure = getTransitiveClosurePerClass(graph, affectedTests);
```
august782: These calls to getTransitiveClosurePerClass are replacing the one that uses classesToAnalyze as an argument. However, so far, it feels as if classesToAnalyze are decomposed into changedClasses and newClasses. Where does affectedTests/allTests fit in? Would there be any problems with calling on just the tests and not all classes as was being done before?

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
```diff
@@ -76,19 +80,28 @@ public Loadables(List<String> classesToAnalyze, String artifactsDir, String sfPa
...
+        } else {
+            transitiveClosure = getTransitiveClosurePerClass(graph, allTests);
+        }
         long transitiveClosureTime = System.currentTimeMillis();
         if (computeUnreached) {
-            unreached = findUnreached(deps, transitiveClosure);
+            //TODO: findUnreached may be affected by incrementally finding test dependencies
+            Map<String, Set<String>> allTransitiveClosure = transitiveClosure;
+            if (incrementalUpdate) {
+                allTransitiveClosure = getTransitiveClosurePerClass(graph, allTests);
```
august782: So no matter what, it is necessary to get the transitive closure based on all tests? Is there any way to do this safely incrementally to get the correct unreached part (I don't think this is very critical though, as transitive closure computation should be fairly fast...)?

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
```diff
@@ -182,11 +196,22 @@ public Loadables create(List<String> moreEdges, Classpath sfClassPath,
             jdepsClassPath = Writer.pathToString(localPaths);
         }
         args.addAll(Arrays.asList("-cp", jdepsClassPath));
-        args.addAll(localPaths);
+        if (incrementalUpdate) {
+            args.addAll(changedClasses);
```
august782: If for an incrementalUpdate there is no need to use localPaths, can that logic be propagated earlier to make sure localPaths doesn't even need to be computed?

starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
```diff
@@ -80,6 +85,34 @@ public static void computeAndSaveNewCheckSums(String artifactsDir,
...
+                if (deps.keySet().contains(left)) {
+                    deps.get(left).add(right);
+                } else {
+                    deps.put(left, new HashSet<>(Collections.singletonList(right)));
+                }
+            }
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+
```
august782: Remove this empty line.

starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
```diff
@@ -93,7 +126,7 @@ public static void computeAndSaveNewCheckSums(String artifactsDir,
             if (deps.keySet().contains(left)) {
                 deps.get(left).add(right);
             } else {
-                deps.put(left, new HashSet<>(Arrays.asList(right)));
+                deps.put(left, new HashSet<>(Collections.singletonList(right)));
```
august782: What's the reasoning for use of the singletonList?

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -38,72 +38,57 @@ public ZLCHelper() {
...
-//        }
-//        long end = System.currentTimeMillis();
-//        System.out.println("[TIME]UPDATING CHECKSUMS: " + (end - start) + "ms");
-//    }
-
-    public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader loader,
-                                     String artifactsDir, Set<String> unreached) {
-        // TODO: Optimize this by only recomputing the checksum+tests for changed classes and newly added tests
+    public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader loader, String artifactsDir,
+                                     Set<String> changed, Set<String> newClasses, boolean incrementalUpdate) {
```
august782: To keep things consistent, shouldn't changed actually be changedClasses?

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -38,72 +38,57 @@ public ZLCHelper() {
...
+                    String extForm = entry.getKey();
+                    if (changed.contains(extForm)) {
+                        String fqn = Writer.urlToFQN(extForm);
+                        Set<String> tests = new HashSet<>();
+                        for (String test : testDeps.keySet()) {
+                            if (testDeps.get(test).contains(fqn)) {
+                                tests.add(test);
+                            }
+                        }
+                        if (tests.isEmpty()) {
```
august782: If tests is empty, is it correct to set the mapping to be empty (that no tests cover this class), or to leave the old dependencies from before? If the former, then this check for isEmpty and then continuing should not be here.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -38,72 +38,57 @@ public ZLCHelper() {
...
+                }
+
+                //handle new data
+                handleNewData(testDeps, loader, newClasses, zlcData);
+            }
+            Writer.writeToFile(zlcData, zlcFile, artifactsDir);
+        }
         long end = System.currentTimeMillis();
-        LOGGER.log(Level.FINE, "[PROFILE] updateForNextRun(updateZLCFile): " + Writer.millsToSeconds(end - start));
+        LOGGER.log(Level.FINE, "[TIME]UPDATING CHECKSUMS: " + Writer.millsToSeconds(end - start));
```
august782: Why is it important to change the logging message? In particular, it might be important to keep the same stuff in the [].

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -113,8 +98,35 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
...
+    public static List<ZLCData> createZLCData(Map<String, Set<String>> testDeps, ClassLoader loader) {
+        long start = System.currentTimeMillis();
+        List<ZLCData> zlcData = new ArrayList<>();
+        Set<String> deps = new HashSet<>();
+        // merge all the deps for all tests into a single set
+        for (String test : testDeps.keySet()) {
+            deps.addAll(testDeps.get(test));
+        }
+
+        // for each dep, find it's url, checksum and tests that depend on it
```
august782: it's -> its

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -113,8 +98,35 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
...
+        handleNewData(testDeps, loader, deps, zlcData);
```
august782: So technically handleNewData() doesn't just handle the new data, it can be used to handle any data (as shown here)? Should the name be changed to reflect this, so it can just be handleData()?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
```diff
@@ -215,10 +218,17 @@ public Result prepareForNextRun(String sfPathString, Classpath sfClassPath, List
         // library jars.
         File libraryFile = new File(jdepsCache, "jdk.graph");
         // Create the Loadables object early so we can use its helpers
-        Loadables loadables = new Loadables(classesToAnalyze, artifactsDir, sfPathString, filterLib, jdepsCache);
+        Loadables loadables = new Loadables(changedClasses, newClasses,
```
august782: Did classesToAnalyze from before contain test classes as well? Is it safe to not include tests as part of the arguments?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
```diff
@@ -215,10 +218,17 @@ public Result prepareForNextRun(String sfPathString, Classpath sfClassPath, List
...
+                artifactsDir, sfPathString, filterLib, jdepsCache);
         // Surefire Classpath object is easier to iterate over without de-constructing
         // sfPathString (which we use in a number of other places)
         loadables.setSurefireClasspath(sfClassPath);
+        // explicitly pass affectedTests to loadables so we can find transitive closure only for affected
+        loadables.setAffectedTests(new ArrayList<>(affectedTests));
+        // for computation of unreached
+        List<String> allTests = new ArrayList<>(affectedTests);
+        allTests.addAll(nonAffected);
```
august782: nonAffected is just non-affected classes, right? It does not include non-test classes?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
```diff
@@ -215,10 +218,17 @@ public Result prepareForNextRun(String sfPathString, Classpath sfClassPath, List
...
         File libraryFile = new File(jdepsCache, "jdk.graph");
         // Create the Loadables object early so we can use its helpers
-        Loadables loadables = new Loadables(classesToAnalyze, artifactsDir, sfPathString, filterLib, jdepsCache);
+        Loadables loadables = new Loadables(changedClasses, newClasses,
+                artifactsDir, sfPathString, filterLib, jdepsCache);
         // Surefire Classpath object is easier to iterate over without de-constructing
         // sfPathString (which we use in a number of other places)
         loadables.setSurefireClasspath(sfClassPath);
+        // explicitly pass affectedTests to loadables so we can find transitive closure only for affected
+        loadables.setAffectedTests(new ArrayList<>(affectedTests));
```
august782: Why are affectedTests and allTests being set explicitly through methods to loadables? Why should they not be set in the constructor, as with changedClasses/newClasses?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -97,16 +106,25 @@ protected void updateForNextRun(Set<String> nonAffected) throws MojoExecutionExc
...
+            newClasses.removeAll(ZLCHelper.getExistingClasses(getArtifactsDir()));
+            // get changed classes in project
+            Set<String> changedClasses = new HashSet<>();
+            changedClasses.addAll(ZLCHelper.getChangedClasses(changedData));
+
             ClassLoader loader = createClassLoader(sfClassPath);
-            //TODO: set this boolean to true only for static reflectionAnalyses with * (border, string, naive)?
-            boolean computeUnreached = true;
-            Result result = prepareForNextRun(sfPathString, sfClassPath, allTests, nonAffected, computeUnreached);
+            Result result = prepareForNextRun(sfPathString, sfClassPath, new ArrayList<>(changedClasses),
```
august782: Why is it important to construct new objects for changedClasses and newClasses but not for affectedTests or nonAffected?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -97,16 +106,25 @@ protected void updateForNextRun(Set<String> nonAffected) throws MojoExecutionExc
...
+                    new ArrayList<>(newClasses), affectedTests, nonAffected, computeUnreached, incrementalUpdate);
+            // add new jar classes for ZLCHelper updateZLCFile
+            newClasses.addAll(Cache.getNewJarClasses());
             Map<String, Set<String>> testDeps = result.getTestDeps();
             graph = result.getGraph();
             Set<String> unreached = computeUnreached ? result.getUnreachedDeps() : new HashSet<String>();
             if (depFormat == DependencyFormat.ZLC) {
-                ZLCHelper zlcHelper = new ZLCHelper();
-                zlcHelper.updateZLCFile(testDeps, loader, getArtifactsDir(), unreached);
+                //ZLCHelper zlcHelper = new ZLCHelper();
```
august782: So before, this was being called and was then clearing out that static map. Was that a bug then?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/SelectMojo.java
```diff
@@ -51,14 +51,15 @@ public void execute() throws MojoExecutionException {
         Set<String> affectedTests = new HashSet<>(allTests);
         Pair<Set<String>, Set<String>> data = computeChangeData();
         Set<String> nonAffectedTests = data == null ? new HashSet<String>() : data.getKey();
+        Set<String> changedClasses  = data == null ? new HashSet<String>() : data.getValue();
```
august782: There's an extra space before " =  data"

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
```diff
@@ -76,19 +80,28 @@ public Loadables(List<String> classesToAnalyze, String artifactsDir, String sfPa
...
-        transitiveClosure = getTransitiveClosurePerClass(graph, classesToAnalyze);
+        if (incrementalUpdate) {
+            transitiveClosure = getTransitiveClosurePerClass(graph, affectedTests);
+        } else {
+            transitiveClosure = getTransitiveClosurePerClass(graph, allTests);
+        }
         long transitiveClosureTime = System.currentTimeMillis();
         if (computeUnreached) {
-            unreached = findUnreached(deps, transitiveClosure);
+            //TODO: findUnreached may be affected by incrementally finding test dependencies
```
denizarsan: Space after `//`

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -113,8 +98,35 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
...
+
+//    public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader loader,
+//                                     String artifactsDir, Set<String> unreached) {
+//        // TODO: Optimize this by only recomputing the checksum+tests for changed classes and newly added tests
+//        long start = System.currentTimeMillis();
+//        List<ZLCData> zlc = createZLCData(testDeps, loader);
+//        Writer.writeToFile(zlc, zlcFile, artifactsDir);
+//        long end = System.currentTimeMillis();
+//        LOGGER.log(Level.FINE, "[PROFILE] updateForNextRun(updateZLCFile): " + Writer.millsToSeconds(end - start));
+//    }
```
denizarsan: I think you should remove this code rather than commenting it out.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -154,7 +166,7 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
                 String newCheckSum = checksumUtil.computeSingleCheckSum(url);
                 if (!newCheckSum.equals(oldCheckSum)) {
                     affected.addAll(tests);
-                    changedClasses.add(stringURL);
+                    changedClasses.add(stringURL);//contain changed/deleted classes and jars
```
denizarsan: Add space after `//` in `//contain..` (I don't know if this can be checked by Checkstyle)

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
```diff
@@ -193,6 +193,7 @@ public Classpath getSureFireClassPath() throws MojoExecutionException {
         long start = System.currentTimeMillis();
         if (sureFireClassPath == null) {
             try {
+                //get All dependencies that this project has, including the project itself
```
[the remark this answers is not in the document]
denizarsan: Space after `//`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -42,6 +43,13 @@
     @Parameter(property = "cleanBytes", defaultValue = "true")
     protected boolean cleanBytes;
 
+    //TODO: set this boolean to true only for static reflectionAnalyses with * (border, string, naive)?
```
denizarsan: Space after `//`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
```diff
@@ -65,15 +73,15 @@ public void execute() throws MojoExecutionException {
...
+            updateForNextRun(nonAffected, changed);
         }
     }
 
     protected Pair<Set<String>, Set<String>> computeChangeData() throws MojoExecutionException {
         long start = System.currentTimeMillis();
         Pair<Set<String>, Set<String>> data = null;
         if (depFormat == DependencyFormat.ZLC) {
-            ZLCHelper zlcHelper = new ZLCHelper();
+            ZLCHelper zlcHelper = new ZLCHelper();// be careful about this. this clears the static zlcDataMap
```
denizarsan: Can you add a more specific comment for this? What should one do or not do? What does "being careful" entail?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
denizarsan: Remove this commented line too.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
```diff
@@ -86,9 +88,12 @@ private void updateForNextRun(List<String> allClasses) throws MojoExecutionExcep
         Classpath sfClassPath = getSureFireClassPath();
         String sfPathString = Writer.pathToString(sfClassPath.getClassPath());
         ClassLoader loader = createClassLoader(sfClassPath);
-        Result result = prepareForNextRun(sfPathString, sfClassPath, allClasses, new HashSet<String>(), false);
-        ZLCHelper zlcHelper = new ZLCHelper();
-        zlcHelper.updateZLCFile(result.getTestDeps(), loader, getArtifactsDir(), new HashSet<String>());
+        List<String> allTests = getTestClasses("updateForNextRun");
+        Result result = prepareForNextRun(sfPathString, sfClassPath, new ArrayList<String>(), allClasses,
+                new HashSet<>(allClasses), new HashSet<String>(), false, incrementalUpdate);
+        //ZLCHelper zlcHelper = new ZLCHelper();// call this would clear zlcDataMap
```
denizarsan: Just remove the commented line.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -63,9 +63,9 @@ public void execute() throws MojoExecutionException {
         Logger.getGlobal().setLoggingLevel(Level.parse(loggingLevel));
         logger = Logger.getGlobal();
         long start = System.currentTimeMillis();
-        setChangedAndNonaffected();
+        setChangedAndNonaffected();//get changed/deleted classes (in project and in jar)
         List<String> excludePaths = Writer.fqnsToExcludePath(nonAffectedTests);
-        setIncludesExcludes();
+        setIncludesExcludes();//set non affected tests in surefire
```
denizarsan: Space after `//`. Alternatively, you can merge these two inline comments into a one block level comment.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -63,9 +63,9 @@ public void execute() throws MojoExecutionException {
         Logger.getGlobal().setLoggingLevel(Level.parse(loggingLevel));
         logger = Logger.getGlobal();
         long start = System.currentTimeMillis();
-        setChangedAndNonaffected();
+        setChangedAndNonaffected();//get changed/deleted classes (in project and in jar)
```
denizarsan: Space after `//`

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
```diff
@@ -91,7 +91,8 @@ protected void run(List<String> excludePaths) throws MojoExecutionException {
         }
         long startUpdateTime = System.currentTimeMillis();
         if (updateRunChecksums) {
-            updateForNextRun(nonAffectedTests);
+            //update dependency graph and zlc mapping for run usage in next revision
```
denizarsan: Space after `//`

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
denizarsan: Can you just remove the commented out code?

denizarsan (changes requested):
Functionally it looks fine to me.

I honestly don't know if we can configure Checkstyle to check for space after `//` but just for the sake of consistency, I tried to mark the places where you were missing one.

starts-core/src/main/java/edu/illinois/starts/helpers/Cache.java
hksskh: already remove this newJarClasses. Since we will run all tests when there are changes in classpath or jar checksums, we actually do not need to track new jar classes in this way.

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
hksskh: The classesToAnalyze passed to Lodables in current STARTS is from prepareForNextRun() in BaseMojo. And the classesToAnalyze passed to this function in BaseMojo is the allTests when calling from DiffMojo and the allClasses when calling from DiffMojo. 
For transitive closure, we only need to find transitive closure for allTests (for DiffMojo) or for allClasses (for impactedmojo), right?
In my modifications, when prepareForNextRun() in BaseMojo is called from DiffMojo, affectedTests and nonaffected tests are passed in, so allTests == classesToAnalyze.
when prepareForNextRun() is called from ImpactedMojo, I passed allClasses as affected and empty list as nonaffected, so the allTests in this case is actually allClasses, so allTests == classesToAnalyze.
Therefore, I think my changes would not cause the problem in your question. However, it truly seems confusing as allTests in this change can be allTests in one case and be allClasses in another case. How do you think I can make this clearer?
Does this long comment makes sense to you? :)

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
hksskh: well I did not take incrementally finding unreached into consideration before! So in order to do this, we need to always make complete transitive closure for allTests, in an incremental way. Then we may need to store transitive closure in some files. However, since find unreached has also been made an option in my changes, I think it is safer to keep find unreached right in this way. Anyway, I may try to look into it after fixing this pull request. :)

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
hksskh: since jdepsClassPath is computed from this localPaths, so I think we may still need to compute this localPaths even in incremental cases.

for the format, in current STARTS the localPaths are actually directory paths to compiled classes and target classes. For the changedClasses and newClasses that I passed in, they are fully qualified names. I have tried on commons-math and jdeps can works with both cases, so I think this should be fine.

starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
hksskh: done.

starts-core/src/main/java/edu/illinois/starts/helpers/RTSUtil.java
hksskh: this is suggested by my Intellij, so I simply refactor it automatically.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed. update tests when it's not empty, and leave it as before when empty.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed. changed function name to addZLCData.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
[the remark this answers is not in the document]
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
hksskh: For the classesToAnalyze before, it is actually allTests if called from DiffMojo, and is allClasses if called from ImpactedMojo. 
Correspondingly, in my changes, the changedClasses and newClasses also include changed classes, if called from DiffMojo.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
hksskh: sorry for these confusions in names. for nonAffected, it is nonaffected tests when called from DiffMojo, and it is nonaffected classes if called from ImpactedMojo. Since this prepareForNextRun() function in BaseMojo is called both from DiffMojo and ImpactedMojo, these parameters can mean differently, so the name seems confusing. How do you think I can  fix this name confusion?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/BaseMojo.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
hksskh: since the parameter type in prepareForNextRun function are Lists for changedClasses and newClasses, but they are actually Sets in DiffMojo, I need to reconstruct Lists from them to match parameter type (this parameter type was set in current STARTS). This seems not efficient, do you think I should change the parameter types in prepareForNextRun function?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
hksskh: yeah this is not a bug before since we do not actually use this static map before, However it causes problems in my changes since I need this static map to incrementally update ZLC file, so I simply remove this line here.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/SelectMojo.java
hksskh: fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/Loadables.java
hksskh: fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed.

hksskh: fixed.

starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/DiffMojo.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
hksskh: fixed.

starts-plugin/src/main/java/edu/illinois/starts/jdeps/RunMojo.java
hksskh: fixed.

hksskh:
@august782 @denizarsan I have add replies to all your comments. Please check them and see if any other changes are needed.

owolabileg:
closing this PR as the requested changes were not addressed since 2017.

PR #31, opened 2017-11-16
Logging README added

jayfurrie:
with "Controlling STARTS Artifact Storage" info


jayfurrie:
apologies Owolabi, I think what I should have done was:

fork from TestingResearchIllinois/starts (master branch)

branch within my own fork and make additions to that branch rather directly to master

pull request for each non-master branch when features are ready

What I did instead was I made changes directly into my forked master branch and pull requested from that. I can see now why that's not how it should be

PR #32, opened 2017-11-16
It windows fix

cptwonton:
fixes integration tests on windows

also adds appveyor integration

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,12 @@
+version: '{build}'
+os: Windows Server 2012
+install:
+  - cmd: mvn --version
+  - cmd: java -version
+build_script:
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+test_script:
+  - mvn verify -B
+cache:
+  - C:\maven\
+  - C:\Users\appveyor\.m2
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,5 +8,4 @@ build_script:
 test_script:
   - mvn verify -B
 cache:
-  - C:\maven\
   - C:\Users\appveyor\.m2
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,12 @@
+version: '{build}'
+os: Windows Server 2012
+install:
+  - cmd: mvn --version
```
august782: Just to confirm, as a followup to discussion in the previous PR, mvn is available on the Appveyor VMs, so there's no need to do any fancy installs, right? I assume these commands are just to tell the VM that it should make mvn (and java) available?

appveyor.yml
```diff
@@ -0,0 +1,12 @@
...
+  - cmd: java -version
+build_script:
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+test_script:
+  - mvn verify -B
+cache:
+  - C:\maven\
```
august782: If Maven is available and no longer needed to be downloaded/installed, then we don't need to cache C:\maven\ anymore, right?

appveyor.yml
cptwonton: This is for reporting purposes to ensure that mvn and java do exist. It adds a few lines to the log so that we know exactly what version of maven and java are being used. It may be unnecessary...

appveyor.yml
cptwonton: yes I think so too, I've removed it

august782:
It just occurred to me, but do we need to do something on the Appveyor website to allow it to build STARTS there, similar to how we have to activate the account for building STARTS on Travis?

cptwonton:
@august782 It will require someone in the TestingResearchTeam organization to create an appveyor account and link it the TestingResearchTeam github to the appveyor account, similar to that of travis.

There is a way around this, by manually specifying webhooks and such, but is easiest if you can directly log in with the github account that the repo is on.

For example, I tried adding testingresearchillinois/starts to my own appveyor, and it was added in a barebones fashion and would not build. Adding my forked starts instead, cptwonton/starts, made everything automagically work.

august782:
I've set up the Appveyor account for this project, so if someone else can double-check this PR (it looks fine to me), we can merge and then see if the Appveyor build will kick off.

PR #33, opened 2017-11-16
Update .travis.yml

cptwonton:
adds build matrix to build on both macOS and linux

adds cache as optimization to avoid redownloading dependencies

removes clean because travis automatically runs mvn install -Dskiptest=true etc... to download dependencies prior to running the script. making the script clean out the previously downloaded dependencies does not make sense

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,17 @@
 language: java
 
-before_install:
-    - jdk_switcher use oraclejdk8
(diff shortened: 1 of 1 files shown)


august782:
Other people may want to comment, but this looks fine to me.

owolabileg:
Thanks @cptwonton :+1:

Issue #36, opened 2017-11-17
travisCI mac builds are slow to start


cptwonton:
detailed here: https://github.com/travis-ci/travis-ci/issues/7304

it seems unavoidable. One of my recent PRs was an "enhancement" to the .travis.yml.
I don't think waiting 20-30 minutes for the macOS build to start is worth it. So I can either:
- remove the macOS portion from the .travis.yml build matrix. the .travis.yml is still enhanced compared to before, it just won't build on macOS
- keep the macOS portion, but set it as optional so that build status will not depend on macOS. In other words, as soon as the linux one finishes, the build is complete. macOS will run when it runs, but build status does not wait or depend on macOS.

Or we can keep things as is. Please let me know which you'd prefer, I can make the change.

The attached screenshot below shows that there was a 15 minute delay. I've noticed delays past 30 minutes before.

![image](https://user-images.githubusercontent.com/3122063/32969828-22a44e0c-cbac-11e7-8dba-cb0f96165d2e.png)

august782:
Thanks for bringing this up. If it is possible to make it optional, as you described, that would be my preferred option. I would appreciate it if you could open a PR that changes the build to make that part optional.

owolabileg:
Yes, I agree with making it optional. Thanks :+1:

owolabileg:
mac builds removed in #39

PR #38, opened 2017-11-17
speeds up travisCI builds

cptwonton:
swaps macOS to be optional, and will not impact build status

As you can see from the below screenshot, the linux build completed 3 minutes ago but the macOS one still hasn't started.
This used to delay the travisCI build status, but now that macOS is considered an optional build the build status reflects the linux one and is passing as soon as the linux one is done.

image


august782:
Just tagging here that this PR relates to issue #36. This PR looks fine to me.

Issue #16, opened 2017-10-05
surefire plugin declaration location in pom.xml of target project


DrRacket:
'mvn starts:starts' failed on NonDex,  reporting 'project nondex: Surefire plugin not available' .
The surefire plugin is declared in the pom.xml of NonDex under \<project>\<build>\<pluginManagement>\<plugins>.
Adding surefire plugin under \<project>\<build>\<plugins> fixes this issue.
The sha for NonDex I'm working with is a1d2fa21e19acd5453abf4c549b61101ac115e3b.

owolabileg:
Fixed in #40

PR #40, opened 2017-11-25
Adding in check for Surefire plugin under pluginManagement, not just the regular build plugins

august782:
This PR adds a check for the Surefire plugin in the pluginManagement section of the project, which is actually fairly common. Addresses issue #16.

### DIFF
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
@@ -77,7 +77,8 @@ public class PomUtil implements StartsConstants {
                 break;
             }
         }
-        // Did not find plugin under the default build plugins, but maybe is under pluginManagement
+        // Did not find plugin under <build><plugins>,
+        // check under <build><pluginManagement><plugins>
         if (plugin == null) {
             plugins = project.getPluginManagement().getPlugins();
             for (Plugin p : plugins) {
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
@@ -77,6 +77,16 @@ public class PomUtil implements StartsConstants {
                 break;
             }
         }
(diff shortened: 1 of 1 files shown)


starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
```diff
@@ -77,6 +77,16 @@ public static Plugin lookupPlugin(String name, MavenProject project) {
                 break;
             }
         }
+        // Did not find plugin under the default build plugins, but maybe is under pluginManagement
```
owolabileg: but maybe is under pluginManagement --> check under pluginManagement

owolabileg: default build plugins --> default <build><plugins>
pluginManagament --> <build><pluginManagement>

starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
```diff
@@ -77,6 +77,16 @@ public static Plugin lookupPlugin(String name, MavenProject project) {
...
+        if (plugin == null) {
+            plugins = project.getPluginManagement().getPlugins();
+            for (Plugin p : plugins) {
```
owolabileg: p --> plugin

owolabileg:
Did you check that this works for NonDex, as reported in #16?

starts-core/src/main/java/edu/illinois/starts/helpers/PomUtil.java
august782: Addressed.

august782: Addressed.

august782:
Yes, we checked that this fix does find the plugin under pluginManagement and utilize starts as expected (no change leads to no tests selected, some changes lead to some tests selected).

owolabileg:
LGTM, merge if all pass

PR #42, opened 2017-12-03
check null obect per EvoSuite test - 2

peifengjing:
Continue modifying for the null bug found by EvoSuite

Past PR: https://github.com/TestingResearchIllinois/starts/pull/27


peifengjing:
This modification change the method in Cache to public List<String> loadM2EdgesFromCache(String pathString)

This will not pass the CI because start-plugin downloads start-core.jar which is the original version of the function public void loadM2EdgesFromCache(List<String> moreEdges, String pathString).

PR #45, opened 2017-12-11
Clean it

cptwonton:
adds integration test for exercising the Clean Mojo
reuses trivial code from First-IT
uses invoker.properties file to first run `starts:starts`, followed by `starts:clean`
finally runs verify.groovy script to check that the .starts directory does not exist.


cptwonton:
removes unnecessary code, IT will specifically test the clean feature and everything else is removed.
test classes are unnecessary since .starts directory is still generated by starts goal and removed by clean goal.
dir renamed to startsDir for more clarify

owolabileg:
Closing this PR as part of the other integration-test related PRs that now seem out of sync. Thanks for the contribution @cptwonton !

PR #46, opened 2017-12-11
second integration-test run will have results of the first integration-test available.

cptwonton:
maven invoker will no longer reset the cloned ITs between integration-test runs. this is necessary for the second STARTS run to build on top of the first STARTS run, otherwise both runs will be fresh and STARTS will generate new checksums each time, meaning all tests will be selected as affected in both runs (that is not what we want and this fixes that).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       jdk: oraclejdk8
 
 script:
-    - mvn verify -B
+    - mvn clean verify -B
 
 cache:
     directories:
--- a/starts-plugin/pom.xml
+++ b/starts-plugin/pom.xml
@@ -65,7 +65,7 @@
           <addTestClassPath>true</addTestClassPath>
           <debug>false</debug>
           <cloneProjectsTo>${project.build.directory}/it</cloneProjectsTo>
-          <cloneClean>true</cloneClean>
+          <cloneClean>false</cloneClean>
           <pomIncludes>
             <pomInclude>*/pom.xml</pomInclude>
           </pomIncludes>
--- a/starts-plugin/src/it/base-it/verify.groovy
+++ b/starts-plugin/src/it/base-it/verify.groovy
@@ -2,16 +2,21 @@
  * Copyright (c) 2015 - Present. The STARTS Team. All Rights Reserved.
  */
 
-import edu.illinois.starts.jdeps.VerifyUtil;
+import edu.illinois.starts.jdeps.VerifyUtil
 
-firstRun = new File(basedir, "first-run.txt");
-verifyUtil = new VerifyUtil(new File(basedir, "build.log"));
+firstRun = new File(basedir, "first-run.txt")
+verifyUtil = new VerifyUtil(new File(basedir, "build.log"))
 
 if (!firstRun.exists()) {
-    firstRun.createNewFile();
-    verifyUtil.assertCorrectlyAffected("1");
+    verifyUtil.assertCorrectlyAffected("1")
+    firstRun.createNewFile()
 } else {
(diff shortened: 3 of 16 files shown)


cptwonton:
From the logs, it seems like the first run of integration-test is all passing but the second pass thru is not.
This is likely due to issues with the setup/verify script. I am looking into this now.

owolabileg:
do we need to change `mvn verify -B` to `mvn clean verify -B` in .travis.yml?

cptwonton:
yes, that will work I think.
The issue is that when we run an integration test the first two times STARTS will work as expected. invoker will not clean up the directory for subsequent runs, meaning we need to do it manually. mvn clean verify will accomplish this. i was looking into adding replaceallinfile method to verifyutil for use in resetting the IT src back to pre-test conditions, but that had various hurdles to jump thru as well. clean is easier.

cptwonton:
but that also means that if we download the code to a machine and run mvn integration-test, it will only work for the first run.
subsequent runs of mvn integration-test will fail unless we clean first... so that's not an ideal solution, right?

owolabileg:
We can either document that `clean` is required or have invoker clean on each run when it is done.

cptwonton:
documenting that clean is required is fine with me.
I think having invoker clean on each run is better, but it would have to be at the beginning - otherwise all the IT artifacts are gone...
What I have now is I added a replaceAllInFile method to VerifyUtil, it does exactly what the replaceAllInFile does in SetupUtil.
I'm using this to revert the code change that happens.
But it's not working, and I think it has to do with how the encoding of the changing Java file starts off as UTF8 w/o BOM, and remains so after setuputil.replaceAllInFile runs on it.
But when verifyutil.replaceAllInFile runs, for some reason it changes the encoding to ANSI
Not entirely sure if this is why the IT is failing. I'm looking specifically at fixing first-it before any of the others.

cptwonton:
yey

cptwonton:
while `SetupUtil` has built in detection to see if it should change the src (presence of zlc file), the `VerifyUtil` does things in a different but equally logical way.
Recent changes (`cloneClean=true`) means we require additional cleanup to do after the second run of the `integration-test` goal. namely, we need to
- revert src code changes back to the original
- delete starts artifacts for subsequent runs of `mvn integration-test`
- remove first-run.txt

These three steps are similar in that all three are done to "reset" the IT back to its original state. They can be logically grouped into the same resetIT() method. The presence of a zlc file like how `SetupUtil` does its detection is not sufficient, as both first and second runs of `integration-test` will have zlc.exists() == true. What does not exist in the first run but will exist in the second run, however, is first-run.txt.
The same is true for SetupUtil - first-run.txt will not exist during the first run of setup.groovy but will exist for the second run of setup.groovy.
So, if the goal is to make logic consistent between both `SetupUtil` and `VerifyUtil`, I claim that first-run.txt.exists() is a better detection method than .zlc.exists(). 

Currently, setup.groovy uses first-run.txt.exists() but in conjunction with the legacy method using zlc.exists().
verify.groovy uses first-run.txt.exist() detection method only.
This also means non-static methods, such as setuputil.replaceAllInFile, can be made static - it will not require member variable access. replaceAllInFile should be consistent between SetupUtil and VerifyUtil.

Please let me know thoughts or if I'm missing something

cptwonton:
@owolabileg on second thought, I'm not sure if this requires squashing/rebasing or not.
whenever we get this PR reviewed and merged, I can rebase all the other new ITs, though.

owolabileg:
I am closing this PR that is now in conflict, we can reopen if someone is still available to work on it. Thanks a lot @cptwonton !!

Issue #25, opened 2017-11-07
use Paths.get() more uniformly


owolabileg:
We need to replace patterns like this: `String path = "a" + File.separator + "b"` with `String path = Path.get("a", "b")` for reasons that were discussed in #13

PR #48, opened 2017-12-12
organizing paths by using a uniform pattern when giving a value

jkhodadadian:
Replaced patterns like `String path = "a" + File.separator + "b"` with `String path = Paths.get("a", "b")` as it was requested in #25


owolabileg:
Closing this PR that is now in conflict; we can reopen if someone is willing to work on it. Thanks @jkhodadadian for the contribution!

PR #49, opened 2017-12-13
Non java it

cptwonton:
building on top of recent IT infrastructure changes, this PR adds 1 additional IT: non-java-IT
using a pre-existing first-it as a template, the non-java-IT will have setup.groovy modify a books.xml file.
the actual first.java does reference the books.xml and actually reads the file in.
the corresponding firsttest.java file includes an additional unit test to ensure that the books.xml file is read in correctly.

the IT will run starts to get a baseline, modify the books.xml file, and run starts again and assert that 0 tests are affected by the change to books.xml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       jdk: oraclejdk8
 
 script:
-    - mvn verify -B
+    - mvn clean verify -B
 
 cache:
     directories:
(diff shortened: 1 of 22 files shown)


owolabileg:
Closing this PR that is now in conflict; we can re-open if someone is available to work on it. Thanks for the contribution @cptwonton !

PR #52, opened 2017-12-13
Is a it

cptwonton:
4 nodes, is-a relationship (extends)


owolabileg:
Closing this PR that is now in conflict; we can re-open if someone is available to work on it. Thanks for the contribution @cptwonton !

PR #53, opened 2017-12-13
Multiple is a it

cptwonton:
seven classes, with mixtures of what extends what.
exercises STARTS ability to construct this graph with 14 nodes and properly detect affected tests


owolabileg:

PR #54, opened 2017-12-14
Diff opt it

cptwonton:
ensures that starts:diff -DupdateDiffChecksums=true successfully updates the checksums. Reuses code+test code from multiple-is-a-it (14 node graph) for this.


owolabileg:

PR #55, opened 2017-12-14
Use a it

cptwonton:
use-a relationship integration test, with 4 nodes.
ensures STARTS can correctly build the graph and find affected tests when we change Second.java between STARTS runs.


cptwonton:
rebased with IT-clone-fix, and updated with
- good coding style
- cleaned up code. IT will specifically test the uses relationship, any extra code is removed

owolabileg:

PR #56, opened 2017-12-15
Parallel solution of off-line mode

yujinz:
Usage: `mvn starts:starts -DwriteNonAffected=true -DofflineMode=true`

This is my final solution of the offline mode of `starts:starts` which runs e phase (test running) and g phase (dependency graph updating) in parallel. 

Since PR #44 has not yet been merged, building the code of this (`parallel`) branch in my repo does not work properly. Please try building the `parallel-before-rebase` branch instead if you would like to see how it works before merging.

### DIFF
--- a/starts-plugin/pom.xml
+++ b/starts-plugin/pom.xml
@@ -26,6 +26,11 @@
       <artifactId>starts-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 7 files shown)


owolabileg:
Closing this PR that is now in conflict; we can re-open if someone is available to work on it. Thanks for the contribution @yujinz !!

PR #69, opened 2021-03-30
Fix NullPointerException in ZLCHelper

jtx1999:
This was causing the current appveyor CI to fail.

### DIFF
--- a/starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
+++ b/starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
@@ -105,8 +105,8 @@ public class ZLCHelper implements StartsConstants {
                 continue;
             }
             URL url = loader.getResource(klas);
-            String extForm = url.toExternalForm();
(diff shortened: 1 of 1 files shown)


starts-core/src/main/java/edu/illinois/starts/helpers/ZLCHelper.java
```diff
@@ -105,8 +105,8 @@ public static void updateZLCFile(Map<String, Set<String>> testDeps, ClassLoader
                 continue;
             }
             URL url = loader.getResource(klas);
-            String extForm = url.toExternalForm();
-            if (url == null || ChecksumUtil.isWellKnownUrl(extForm) || (!useJars && extForm.startsWith("jar:"))) {
+            if (url == null || ChecksumUtil.isWellKnownUrl(url.toExternalForm())
```
jtx1999, line 108: Add extra if statement and cache url.toExternalForm

jtx1999: Done

jtx1999:
@owolabileg Please review

PR #85, opened 2021-05-28
Fix ZLC format

jtx1999:
- Set `cloneClean` to false so the integration tests can test starts on modified code, thanks to @cptwonton
 - Run the tests on after `mvn clean` on AppVeyor and Travis, since the integration tests won't auto overwrite now.
 - Fix indexing issue in reading PLAIN_TEXT deps.zlc file, as we should read from the 0th line.
 - Comment out integration test on CLZ format. CLZ is not working currently. See #86 
 - Fix casting in EkstaziHelper.java. `toArray()` returns `Object[]` and `toArray(new String[])` returns `String[]`.
   - Was getting error on Java 9+: java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Ljava.lang.String; ([Ljava.lang.Object; and [Ljava.lang.String; are in module java.base of loader 'bootstrap')
 - Fix setting excludes in SurefireMojoInterceptor.java on Windows.
   - On Windows, STARTS get non-affected test paths with backward-slash separator. Surefire only accept excludes with forward-slash. So all tests were executed previously.
   - Replacing the file separator to forward slash to exclude correct tests.


jtx1999:
@owolabileg Please review. Those fixes should really be in separate PRs, but most of them are discovered while I was trying to fix the ZLC format.

PR #91, opened 2022-06-20
Convert local variables to fields for access

thenewpyjiang:


starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
```diff
@@ -31,6 +32,13 @@
 @Mojo(name = "impacted", requiresDirectInvocation = true, requiresDependencyResolution = ResolutionScope.TEST)
 @Execute(phase = LifecyclePhase.TEST_COMPILE)
 public class ImpactedMojo extends DiffMojo implements StartsConstants {
+
+    protected Set<String> impacted;
+    protected Set<String> nonAffected;
+    protected Set<String> changed;
+    protected Set<String> newClasses;
+    protected Set<String> oldClasses;
```
owolabileg, line 40: make these fields private

owolabileg (changes requested):

PR #93, opened 2022-07-07
Add getter and setter for updateImpactedChecksums

thenewpyjiang:
Add getter and setter for updateImpactedChecksums, so that eMOP can make use of them.

### DIFF
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
@@ -96,10 +96,6 @@ public class ImpactedMojo extends DiffMojo implements StartsConstants {
         return Collections.unmodifiableSet(oldClasses);
     }
 
-    public boolean isUpdateImpactedChecksums() {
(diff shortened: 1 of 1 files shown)


starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
```diff
@@ -96,6 +96,14 @@ public Set<String> getOldClasses() {
         return Collections.unmodifiableSet(oldClasses);
     }
 
+    public boolean getUpdateImpactedChecksums() {
```
owolabileg, line 99: please start getter for a booleans with `is`, not `get`

owolabileg (changes requested):

starts-plugin/src/main/java/edu/illinois/starts/jdeps/ImpactedMojo.java
owolabileg: You are right: the phrasing is weird. May I ask that you revert to what you had? Also, do you really need a getter for the feature that you are implementing?

PR #110, opened 2023-10-28
Optimize method and hybrid starts by removing unecessary affected test classes computation

mmk-1:
- I've removed the parts that computed `affectedTestClasses` for optimization.
- `nonAffectedMethods` variable was not used anyways in the code so I removed that too
- As an outcome of removing `affectedTestClasses` computations, `methodToTestClasses` mapping wasn't needed in `HybridMojo.java` either.

### DIFF
--- a/starts-plugin/src/main/java/edu/illinois/starts/jdeps/HybridMojo.java
+++ b/starts-plugin/src/main/java/edu/illinois/starts/jdeps/HybridMojo.java
@@ -63,6 +63,12 @@ public class HybridMojo extends DiffMojo {
     @Parameter(property = "debug", defaultValue = TRUE)
     private boolean debug;
 
+    /**
+     * Set this to "true" to compute affected test classes as well.
+     */
+    @Parameter(property = "computeAffectedTests", defaultValue = FALSE)
+    private boolean computeAffectedTests;
+
     public void setComputeImpactedMethods(boolean computeImpactedMethods) {
         this.computeImpactedMethods = computeImpactedMethods;
     }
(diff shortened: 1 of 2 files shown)


starts-plugin/src/main/java/edu/illinois/starts/jdeps/HybridMojo.java
```diff
@@ -160,7 +153,6 @@ public void execute() throws MojoExecutionException {
         // Build method level static dependencies
         try {
             MethodLevelStaticDepsBuilder.buildMethodsGraph(includeVariables);
-            methodToTestClasses = MethodLevelStaticDepsBuilder.computeMethodToTestClasses();
```
owolabileg, line 163: I don't understand this change: what was it doing before?

owolabileg (changes requested):

owolabileg:
@mmk-1 let's think carefully about these changes, before merging them.

> I've removed the parts that computed affectedTestClasses for optimization.

IIUC, these affectedTestClasses maps method-level changes to the class level. Is that right? If so, we'd need something like that for experiments. So, instead of deleting, why not make it an option that the user can enable or disable?

> nonAffectedMethods variable was not used anyways in the code so I removed that too

It seems that you would need this variable when you later implement all the variants (e.g., to avoid instrumenting nonAffected methods. Does that make sense?

> As an outcome of removing affectedTestClasses computations, methodToTestClasses mapping wasn't needed in HybridMojo.java either.

Please see my comment above: should we make it an option instead of outright deleting it?

starts-plugin/src/main/java/edu/illinois/starts/jdeps/HybridMojo.java
mmk-1: This mapped methods to their dependent test classes. It was used before for finding affectedTestClasses

mmk-1:
Yes you're right. I'll make changes to introduce a parameter to turn this feature on/off

mmk-1:
@owolabileg I have added new parameters based on your reviews to make Affected Test Class computation optional. Please review. Thanks!

owolabileg (approved):

owolabileg:
This all looks good to me, and I think you can merge.

